PR #81, opened 2017-10-09
read user settings from vscode and update them  to java debugger.

andxu:


package.json
```diff
@@ -188,6 +188,26 @@
...
+        },
+        "java.debug.showHex": {
+          "type": "boolean",
+          "default": false
+        },
+        "java.debug.showStaticVariables": {
+          "type": "boolean",
+          "default": false
+        },
+        "java.debug.showQualifiedNames": {
```
yaohaizh: Should be showFullyQualifiedName

yaohaizh: And should add description for the user setttings

src/extension.ts
```diff
@@ -11,20 +11,29 @@ const status: any = {};
...
             status.debugging = "startDebugSession";
 
             try {
-                try {
-                    const level = await configLogLevel(vscode.workspace.getConfiguration().get("java.debug.logLevel"));
-                    console.log("setting log level to ", level);
-                } catch (err) {
-                    // log a warning message and continue, since logger failure should not block debug session
-                    console.log("Cannot set log level to java debuggeer.")
+                if (configurationDirty) {
```
yaohaizh: Extract this as a method to clear the logic

src/extension.ts
```diff
@@ -11,20 +11,29 @@ const status: any = {};
 export function activate(context: vscode.ExtensionContext) {
     // The reporter will be initialized by the later telemetry handler.
     let reporter: TelemetryReporter = null;
-
+    let configurationDirty: boolean = true;
```
yaohaizh: Naming:  configurationDirty  ==> isConfigurationDirty

Also, the dirty is not appropriate here. Maybe just `update` is oaky

src/extension.ts
```diff
@@ -164,3 +181,27 @@ function convertLogLevel(commonLogLevel: string) {
             return "FINE";
     }
 }
+
+function collectUserSettings(): string[] {
+    const settings: string[] =  [];
+    if (vscode.workspace.getConfiguration().get("java.debug.settings.showHex")) {
+        settings.push("--show_hex");
```
yaohaizh: Is the `--` necessary in here. This is not console parameter. In this case, we should just naming setting name for setting values.

README.md
```diff
@@ -59,6 +59,11 @@ Please also check the documentation of [Language Support for Java by Red Hat](ht
 ### User Settings
 
 - `java.debug.logLevel`: minimum level of debugger logs that are sent to VS Code, defaults to `warn`.
+- `java.debug.settings.showHex`: whether or not the number should be displayed as hex format in variable view, defaults to `false`.
+- `java.debug.settings.showStaticVariables`: whether or not the static variables will be displayed in variable view, defaults to `false`.
+- `java.debug.settings.showQualifiedNames`: wheter or not the class name shoulde be displayed using the fully qualified class name, defaults to `false`.
```
testforstephen: typos. `shoulde` -> `should`

src/extension.ts
```diff
@@ -164,3 +176,43 @@ function convertLogLevel(commonLogLevel: string) {
             return "FINE";
     }
 }
+
+function collectUserSettings(): string[] {
```
yaohaizh: I'm curious why not just use the plain object but use the key map here. The key map is error-prone since it doesn't provide the strong type verification and spell check. And it is tedious to encode/decode the key/values  in the client/server side.

Issue #82, opened 2017-10-09
Update with the new debug authoring from VSCode 1.17.0


yaohaizh:
https://code.visualstudio.com/updates/v1_17#_multi-root-guide-for-extension-authors

PR #87, opened 2017-10-11
fix feature break from October VSCode insider

testforstephen:

#82

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -5,9 +5,7 @@ import * as vscode from "vscode";
 import TelemetryReporter from "vscode-extension-telemetry";
 import * as commands from "./commands";
 
-export class JavaConfigurationProvider implements vscode.DebugConfigurationProvider {
-    private _resolving: boolean = false;
-
+export class JavaDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
     constructor(private _reporter: TelemetryReporter) {
     }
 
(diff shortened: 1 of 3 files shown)


src/configurationProvider.ts
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import TelemetryReporter from "vscode-extension-telemetry";
+import * as commands from "./commands";
+
+export class JavaConfigurationProvider implements vscode.DebugConfigurationProvider {
```
yaohaizh: JavaDebugConfigurationProvider

src/configurationProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+            type: "java",
+            name: "Debug (Attach)",
+            request: "attach",
+            hostName: "localhost",
+            port: 0,
+        }];
+    }
+
+    // Try to add all missing attributes to the debug configuration being launched.
+    public resolveDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken):
```
yaohaizh: This method is not marked async yet call async meethods

testforstephen: This is VSCode API built-in methods. No need to change it to async. because the result `vscode.ProviderResult<vscode.DebugConfiguration>` could be T or Thenable<T>.

src/configurationProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+                return undefined;
+            } else if (config.request === "launch") {
+                if (!config.mainClass) {
+                    vscode.window.showErrorMessage("Please specify the mainClass in the launch.json.");
+                    return undefined;
+                } else if (!config.classPaths || !Array.isArray(config.classPaths) || !config.classPaths.length) {
+                    config.classPaths = await resolveClasspath(config.mainClass, config.projectName);
+                }
+                if (!config.classPaths || !Array.isArray(config.classPaths) || !config.classPaths.length) {
+                    vscode.window.showErrorMessage("Cannot resolve the classpaths automatically, please specify the value in the launch.json.");
```
andxu: log error?

src/configurationProvider.ts
yaohaizh: That's the point. If you mark as async, then the caller will know how to deal with it. If you don't provide the same behavior as method signature, the behavior will be undefined in some cases.

PR #97, opened 2017-10-16
Refactor attach config parameters

testforstephen:


akaroml:
As discussed, we don't need to change the entry names but improve the documentation if that applies.

PR #108, opened 2017-10-24
Add ability customize resolved classpaths

### LINKED ISSUE #93: Support set extra classpath with maven project
I need to add an extra classpath based on a maven project.
If I add the extra classpath to `classPaths`, the main class can't be found.

It seems launch option `classPaths` will overwrite classpaths automatically resolved from maven project.

Dose the debugger support set extra classpath and preserve automatically resolved classpaths just like eclipse?

supersonicclay:
Adds the following to `launch.json`:

- `classPathsToAdd` - Additional classpaths to add before launching the JVM.
- `classPathsToRemove` - Classpaths to remove before when launching the JVM.

Fixes #93


testforstephen:
@claycephas thanks for the PR. A little concern about the option `classPathsToRemove`, can you list some scenarios that needs this requirement?

README.md
```diff
@@ -45,6 +45,8 @@ Please also check the documentation of [Language Support for Java by Red Hat](ht
 - `args` - The command line arguments passed to the program.
 - `sourcePaths` - The extra source directories of the program. The debugger looks for source code from project settings by default. This option allows the debugger to look for source code in extra directories.
 - `classPaths` - The classpaths for launching the JVM. If not specified, the debugger will automatically resolve from current project.
+- `classPathsToAdd` - Additional classpaths to add before launching the JVM.
+- `classPathsToRemove` - Classpaths to remove before when launching the JVM.
```
testforstephen: A little concern about this option. Can you list some scenarios that needs this option?

claycephas:
When debugging my maven-based project with VS Code, it is including test classes in the classpath. However, IntelliJ and Eclipse don't include those test classes in the classpath. I thought the `classPathsToRemove` could be a temporary workaround for issues like this. I see it as an override so that users have the ability to have some control but without having to list every single class path in the `classPaths` config setting.

testforstephen:
Including test classes in auto resolved maven classpaths looks like a bug. I open a bug for this. https://github.com/Microsoft/vscode-java-debug/issues/111

And the existing `classPaths` option is always overriding the auto resolved classpaths, that's not good. 
I'm thinking whether it's necessary to expose multiple classpaths options to user because we want to keep the configurations as simple as possible. Three classpaths options could confuse the users.

if we change the current `classPaths` behavior to merge instead of overriding, do you think is that OK?

JopBrown:
How about add a flag option to turn auto resolved true/false and `classPaths` option always merge with resolved classpaths.
If the flag is false, resolved classpaths is empty, `classPaths` option behavior just like override.
The flag default to true if ignore, less confusing.

testforstephen:
@claycephas @JopBrown Your suggestion could be a choice. And i also find a new choice to keep compatible with current behavior.

Reference to `New Debug Configuration` action in eclipse, it will persist the auto resolved class paths in configuration file, and user can add/remove the class paths there. Maybe we could take the similar policy. We could let the debugger to save the auto resolved classpaths when new a launch.json file, then the user could add/remove the classpaths in launch.json directly.

JopBrown:
@testforstephen
If the project setting change, will persisted class paths in `launch.json` auto update too?
And I think leave a long list of resolved class paths in `launch.json` is more confuse. 
As an user, I don't care what auto resolved class actually is, just keep sync with maven, and be able to add/remove custom class paths.
It will be complicated for me to update `launch.json` everytime when project setting change.

claycephas:
@testforstephen Regarding placing generated classpaths in `launch.json`:
There's already a generated `.classpath` file. I'd be hesitant to duplicate the classpaths in the `.vscode/launch.json` file. That'd mean that dependencies would effectively be duplicated in `pom.xml`, `.classpath`, and `launch.json`. Typically `launch.json` is source-controlled and typically generated code is not placed in source control.

@JopBrown Regarding your suggestion, there'd be no way to remove a specific problematic classpath (e.g. like `test-classes`) without having to list all the rest (which is over 200 in my case).

The thing I like about `classPathsToAdd` and `classPathsToRemove` is that they're explicit in their intentions and allow lots of flexibility to users when they need a workaround for cases like #111.

testforstephen:
@claycephas We're supporting java 9, it uses a new option `module-path`, let's hold on this PR first until we get a reasonable solution about customizing modulePaths/classPaths.

testforstephen:
This PR is replaced by PR https://github.com/microsoft/vscode-java-debug/pull/1005, which uses the predefined variables in classPaths/modulePaths of launch.json to customize the classpaths.

Here is the predefined variables:
- $Auto - Automatically resolve the classpaths of current project
- $Runtime - The classpaths within 'runtime' scope of current project
- $Test - The classpaths within 'test' scope of current project
- !\<path\> - Exclude the specified path from classpaths.

You can see the sample below to add/remove classpath to launch.json.

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "java",
            "name": "Launch Foo",
            "request": "launch",
            "mainClass": "foo.bar.Foo",
            "classPaths": [
                "$Auto",
                "<extra class path>",
                "!<excluded class path>
            ]
        }
    ]
}

```

PR #141, opened 2017-11-08
fix regression: launch.json isn't generated for project

ansyral:
Signed-off-by: xuzho <xuzho@microsoft.com>

there are two options for files that are opened in a folder but don't have project:

1. generate  a launch.json
2. not generate a launch.json

right now i implemented as option2, what do you think about this?

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -5,6 +5,9 @@ import * as vscode from "vscode";
 import TelemetryReporter from "vscode-extension-telemetry";
 import * as commands from "./commands";
 
+import fs = require("fs");
+import path = require("path");
+
 export class JavaDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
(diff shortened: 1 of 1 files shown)


src/configurationProvider.ts
```diff
@@ -5,6 +5,9 @@ import * as vscode from "vscode";
 import TelemetryReporter from "vscode-extension-telemetry";
 import * as commands from "./commands";
 
+import fs = require("fs");
```
yaohaizh:
```js
import * as fs from "fs";
import * as path from "path";
```

src/configurationProvider.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as fs from "fs";
+import * as path from "path";
```
testforstephen: remove unused imports.

PR #143, opened 2017-11-09
Disable stopOnEntry by default

testforstephen:


ansyral:
also updates package.json?

testforstephen:
@ansyral  No need to change package.json. The default value in package.json is used as code snippet.

PR #176, opened 2017-11-24
only do rebuild for single file

ansyral:
Signed-off-by: xuzho <xuzho@microsoft.com>

1. rebuild only for single file
2. fix regression of multi-root
3. for single file, if its build fails, would ask user to choose whether to proceed. currently, if abort the launch for single file, vscode would pop up an error message "cannot read property openconfigfile of undefined", i've opened an [issue](https://github.com/Microsoft/vscode/issues/39080) to track

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -34,7 +34,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return vscode.window.withProgress({location: vscode.ProgressLocation.Window}, (p) => {
             return new Promise((resolve, reject) => {
                 p.report({message: "Auto generating configuration..."});
-                resolveMainClass(folder.uri).then((res: any[]) => {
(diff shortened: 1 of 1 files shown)


src/configurationProvider.ts
```diff
@@ -34,7 +34,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return vscode.window.withProgress({location: vscode.ProgressLocation.Window}, (p) => {
             return new Promise((resolve, reject) => {
                 p.report({message: "Auto generating configuration..."});
-                resolveMainClass(folder.uri).then((res: any[]) => {
```
yaohaizh: In which case the folder in empty? Single file? 

Does the java-debug also should be changed for this changes?

ansyral: if client only opens a file, the folder is `undefined`. java-debug has handled this.

ansyral:
verified the [issue](https://github.com/Microsoft/vscode/issues/39080) is fixed in vscode insider

PR #189, opened 2017-12-04
Add HCR document update event.

yaohaizh:
@akaroml @testforstephen @andxu @ansyral

### DIFF
--- a/src/hotCodeReplace.ts
+++ b/src/hotCodeReplace.ts
@@ -1,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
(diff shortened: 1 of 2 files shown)


src/hotCodeReplace.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+
+const runningSessions: Set<vscode.DebugSession> = new Set();
```
testforstephen: Find no one is using `runningSessions`. What's the purpose of this cache?

yaohaizh: Some code is removed in this review for custom events.

PR #195, opened 2017-12-15
Add HCR document update event.

yaohaizh:
@akaroml @testforstephen @andxu

### DIFF
--- a/src/hotCodeReplace.ts
+++ b/src/hotCodeReplace.ts
@@ -9,17 +9,23 @@ const suppressedReasons: Set<string> = new Set();
 
 const NOT_SHOW_AGAIN: string = "Not show again";
 
+const JAVA_LANGID: string = "java";
+
+const HCR_EVENT = "hotCodeReplace";
+
+const SAVEDOCUMENT_EVENT = "saveDocument";
+
 export function startHotCodeReplace(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.debug.onDidStartDebugSession((session) => {
         const t = session ? session.type : undefined;
-        if (t === "java") {
+        if (t === JAVA_LANGID) {
             runningSessions.add(session);
         }
     }));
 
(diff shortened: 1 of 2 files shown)


src/hotCodeReplace.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (t === "java") {
+            runningSessions.delete(session);
+            suppressedReasons.clear();
+        }
+    }));
+
+    context.subscriptions.push(vscode.debug.onDidReceiveDebugSessionCustomEvent((customEvent) => {
+        const t = customEvent.session ? customEvent.session.type : undefined;
+        if (t === "java") {
+            if (customEvent.event === "hotCodeReplace" && !suppressedReasons.has(customEvent.body.message)) {
```
akaroml: The event name should be hotCodeReplaceEnd.
And we also need the hotCodeReplaceStart event to let user know the progress.

src/hotCodeReplace.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const runningSessions: Set<vscode.DebugSession> = new Set();
+
+const suppressedReasons: Set<string> = new Set();
+
+const NOT_SHOW_AGAIN: string = "Not show again";
+
+const JAVA_LANGID: string = "java";
+
+const HCR_EVENT = "hotCodeReplace";
```
akaroml: I think we need two events: hotCodeReplaceStart and hotCodeReplaceEnd. And for any event, we need the error object to see whether it's a successful one.

hotCodeReplaceStart will be used to let user know that we are in the progress of redefining classes.

yaohaizh: Okay, we need refine the event types for HCR.  

Currently, this event current only for error handling on the user side.

For hotCodeReplacestart/hotCodeReplaceEnd we need first add them on the debug server side.

yaohaizh: Refined HCR event.

PR #244, opened 2018-03-05
Refine failure scenario.

yaohaizh:
@akaroml @testforstephen @andxu

### DIFF
--- a/src/hotCodeReplace.ts
+++ b/src/hotCodeReplace.ts
@@ -5,7 +5,11 @@ import * as vscode from "vscode";
 
 const suppressedReasons: Set<string> = new Set();
 
-const NOT_SHOW_AGAIN: string = "Not show again";
+const YES_BUTTON: string = "Yes";
+
+const NO_BUTTON: string = "No";
+
+const NEVER_BUTTON: string = "Not show again";
 
 const JAVA_LANGID: string = "java";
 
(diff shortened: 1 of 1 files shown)


src/hotCodeReplace.ts
```diff
@@ -34,19 +38,23 @@ export function initializeHotCodeReplace(context: vscode.ExtensionContext) {
...
             if (!suppressedReasons.has(customEvent.body.message)) {
-                vscode.window.showInformationMessage(`Hot code replace failed - ${customEvent.body.message}`, NOT_SHOW_AGAIN).then((res) => {
-                    if (res === NOT_SHOW_AGAIN) {
-                        suppressedReasons.add(customEvent.body.message);
-                    }
-                });
+                vscode.window.showInformationMessage(
+                    `Hot code replace failed - ${customEvent.body.message}. Would you like to restart the debug session?`,
+                    YES_BUTTON, NO_BUTTON, NEVER_BUTTON).then((res) => {
+                        if (res === NEVER_BUTTON) {
```
testforstephen: What does `Never` mean?

testforstephen: Shouldn't we always show the error message when `HCR` failed? Refer to `Eclipse`, can it be `suppressed`?

yaohaizh: We are not exactly match eclipse behavior.  Eclipse can configurate on its settings UI>

yaohaizh: Not show again for NEVER

PR #262, opened 2018-03-26
fix issue 187:If only 1 main class detected, proceed and start it dir…

andxu:
…ectly

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -240,8 +240,8 @@ function convertLogLevel(commonLogLevel: string) {
 }
 
 interface IMainClassOption {
-    readonly projectName?: string;
-    readonly mainClass?: string;
+    readonly projectName: string;
+    readonly mainClass: string;
 }
 
 async function chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption> {
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -34,7 +34,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return vscode.window.withProgress({location: vscode.ProgressLocation.Window}, (p) => {
             return new Promise((resolve, reject) => {
                 p.report({message: "Auto generating configuration..."});
-                resolveMainClass(folder ? folder.uri : undefined).then((res: any[]) => {
+                resolveMainClass(folder ? folder.uri : undefined).then((res: IMainClassOption []) => {
                     let cache;
                     cache = {};
                     const launchConfigs = res.map((item) => {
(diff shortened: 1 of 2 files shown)


src/configurationProvider.ts
```diff
@@ -134,7 +134,9 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     }).sort ((a, b): number => {
                         return a.label > b.label ? 1 : -1;
                     });
-                    const selection = await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" });
+                    const selection = pickItems.length > 1 ?
+                        await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" })
```
yaohaizh: Select main class<project name> ?? Not understand which should be selected? Both?

src/configurationProvider.ts
```diff
@@ -134,7 +134,9 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     }).sort ((a, b): number => {
                         return a.label > b.label ? 1 : -1;
                     });
-                    const selection = await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" });
+                    const selection = pickItems.length > 1 ?
+                        await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" })
+                        : pickItems[0];
```
testforstephen: If there is no main class in the workspace, what happens here?

andxu: early exit at the top of this method.

src/configurationProvider.ts
andxu: yes, if it is not specified in launch.json, java debug will list them for the choosing from user, if only one main class is found, I will use it directly without popup windows for one chosen option.

src/configurationProvider.ts
testforstephen: Please extract the logic to a method, e.g. `promptForMainClass`

src/configurationProvider.ts
```diff
@@ -262,3 +238,43 @@ function convertLogLevel(commonLogLevel: string) {
...
     }
 }
+
+interface IMainClassOption {
+    readonly projectName?: string;
+    readonly mainClass?: string;
+}
+
+async function chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption> {
+    const res = <any[]>(await resolveMainClass(folder ? folder.uri : undefined));
```
testforstephen: `any[]` => `IMainClassOption[]`

src/configurationProvider.ts
```diff
@@ -262,3 +238,43 @@ function convertLogLevel(commonLogLevel: string) {
             return "FINE";
     }
 }
+
+interface IMainClassOption {
+    readonly projectName?: string;
+    readonly mainClass?: string;
```
testforstephen: Can `mainClass` be empty?

andxu: as a model, I don't want to forbid mainClass to null

PR #267, opened 2018-03-30
doesn't popup error message when user cancels the main class selection.

andxu:


src/configurationProvider.ts
```diff
@@ -189,6 +189,44 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
     private isEmptyArray(configItems: any): boolean {
         return !Array.isArray(configItems) || !configItems.length;
     }
+
+    private async chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption> {
```
yaohaizh: Why the folder can be undefined?  It is strange to declare a type of undefined? If it is not existing, just null value.

testforstephen: Declare the return type to `Promise<IMainClassOption | undefined>`

src/configurationProvider.ts
```diff
@@ -189,6 +189,44 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        }
+        const selection = pickItems.length > 1 ?
+            await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" })
+            : pickItems[0];
+        if (selection && selection.item) {
+            return selection.item;
+        } else {
+            vscode.window.showErrorMessage("Please specify the mainClass (e.g. [mymodule/]com.xyz.MainClass) in the launch.json.");
+            this.log("usageError", "Please specify the mainClass (e.g. [mymodule/]com.xyz.MainClass) in the launch.json.");
+            return undefined;
```
testforstephen: Given that you have used `return ` in the previous side branch, use `return` directly here to keep code consistent.

src/configurationProvider.ts
```diff
@@ -189,6 +189,44 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+            return {
+                description: details,
+                label: name,
+                item,
+            };
+        }).sort ((a, b): number => {
+            return a.label > b.label ? 1 : -1;
+        });
+        if (pickItems.length === 0) {
+            // user gives up the selection, quit sliently
```
testforstephen: typos. `silently`.

src/configurationProvider.ts
```diff
@@ -189,6 +189,44 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
```
yaohaizh: There is no items. User doesn't have the selection yet. Should remove this comment?

PR #287, opened 2018-04-24
auto select current file as the main class if it contains the mainClass.

andxu:


src/configurationProvider.ts
```diff
@@ -212,6 +212,21 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        if (vscode.window.activeTextEditor
+            && vscode.window.activeTextEditor.document && vscode.window.activeTextEditor.document.fileName) {
+                const narrowedChoice = pickItems.filter((options) => {
+                    return options && options.item && options.item.filePath &&
+                        (options.item.filePath === vscode.window.activeTextEditor.document.fileName ||
+                            (process.platform === "win32"
+                                && options.item.filePath.toLowerCase() === vscode.window.activeTextEditor.document.fileName.toLowerCase()));
+                });
+
+                if (narrowedChoice.length === 1) {
```
akaroml: Instead of remove items from the list, can we highlight the item that is currently selected from the active document?

andxu: Not available in vscode

PR #317, opened 2018-06-26
fix vulnerabilities

Eskibear:
as suggested by `npm audit`

### DIFF
--- a/package.json
+++ b/package.json
@@ -372,13 +372,13 @@
     "@types/mocha": "^2.2.32",
     "@types/node": "^6.0.40",
     "cross-env": "^5.2.0",
-    "gulp": "^3.9.1",
+    "gulp": "^4.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -372,13 +372,13 @@
     "@types/mocha": "^2.2.32",
     "@types/node": "^6.0.40",
     "cross-env": "^5.2.0",
-    "gulp": "^3.9.1",
+    "gulp": "^4.0.0",
```
yaohaizh: Be careful with gulp. 4.0.0 and v3.9.1 are totally different.

Eskibear: yes, high risks along with the major version change.

Eskibear:
Potential breaking changes in new major version of `mocha` and `gulp`. It would be safer to wait for their fixes in minor/patch version, instead of forcing to upgrade.

Issue #355, opened 2018-08-07
Add validation for configuration items of launch.json


testforstephen:
The debugger only has type validation for the config items (mainClass, projectName) of launch.json, this is a kind of static validation, and it cannot check the illegal value. 

For example, for the common config errors below, the debugger will report it at a much latter stage with error message `Failed to resolve classpath: ...`. It doesn't point out the failure reason clearly.
- Main class 'C:\demo\org\microsoft\app\Main.java' doesn't exist in the workspace.
- Main class 'org.microsoft.app.Main' isn't unique in the workspace.
- The project 'demo' is not a valid java project.

 In order to better remind the user about the config error with more concise error message, plan to add code validation in the debug configuration provider for the key configuration items. (e.g. mainClass, projectName, port)

testforstephen:
So far VSCode doesn't support auto complete when the user is editing the launch.json. It provides another mechanism `DebugConfigurationProvider` to allow the debugger to resolve launch config before launching debug session.

The idea is to validate mainClass and projectName config in the `DebugConfigurationProvider`, and when the user configured an incorrect value for these two configs, the debugger will prompt the possible correct configs with QuickPick dropdown. If the user selects the fix, the debugger will persist the fix in launch.json too.

testforstephen:
When showing the quick fix proposals, put the proposals with the same as the original mainClass or projectName configs in launch.json on the top, and the remaining proposals have alphabet order.

testforstephen:
PR https://github.com/Microsoft/vscode-java-debug/pull/368 and https://github.com/Microsoft/java-debug/pull/205 merged. It will be released in 0.12.0. close it.

PR #368, opened 2018-08-15
Validate mainClass and projectName before launch

testforstephen:

See issue https://github.com/Microsoft/vscode-java-debug/issues/355 for the requirement. 

The validation work is done on java code side, see the peer PR: https://github.com/Microsoft/java-debug/pull/205

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -118,15 +118,14 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                         return undefined;
                     }
                 }
-                if (!config.mainClass) {
-                    const userSelection = await this.chooseMainClass(folder);
-                    if (!userSelection || !userSelection.mainClass) {
-                        // the error is handled inside chooseMainClass
-                        return;
-                    }
-                    config.mainClass = userSelection.mainClass;
-                    config.projectName = userSelection.projectName;
+
+                const userSelection = await this.quickFixLaunchConfig(folder ? folder.uri : undefined, config);
+                if (!userSelection || !userSelection.mainClass) {
+                    return;
                 }
+                config.mainClass = userSelection.mainClass;
+                config.projectName = userSelection.projectName;
+
                 if (this.isEmptyArray(config.classPaths) && this.isEmptyArray(config.modulePaths)) {
                     const result = <any[]>(await resolveClasspath(config.mainClass, config.projectName));
                     config.modulePaths = result[0];
@@ -195,8 +194,101 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return !Array.isArray(configItems) || !configItems.length;
     }
 
-    private async chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption | undefined> {
-        const res = await resolveMainClass(folder ? folder.uri : undefined);
+    private async quickFixLaunchConfig(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration):
+        Promise<IMainClassOption | undefined> {
+        if (!config.mainClass) {
+            return this.chooseMainClass(folder);
+        }
+
+        const validationResponse = await validateLaunchConfig(folder, config.mainClass, config.projectName);
+        if (validationResponse.mainClass.passed && validationResponse.projectName.passed) {
+            return {
+                mainClass: config.mainClass,
+                projectName: config.projectName,
+            };
+        } else {
+            const errors: string[] = [];
+            if (!validationResponse.mainClass.passed) {
+                errors.push(String(validationResponse.mainClass.message));
+            }
+
+            if (!validationResponse.projectName.passed) {
+                errors.push(String(validationResponse.projectName.message));
+            }
+
+            if (validationResponse.proposals && validationResponse.proposals.length) {
+                const answer = await utility.showErrorMessageWithTroubleshooting({
+                    message: errors.join("\n"),
+                    type: Type.USAGEERROR,
+                }, "Quick Fix");
+                if (answer === "Quick Fix") {
+                    const quickFix: IMainClassOption = await this.showMainClassQuickPick(validationResponse.proposals,
+                        "Please select main class<project name>", true);
+                    this.persistLaunchConfig(folder, config, quickFix);
+                    return quickFix;
+                }
+            } else {
+                utility.showErrorMessageWithTroubleshooting({
+                    message: errors.join("\n"),
+                    type: Type.USAGEERROR,
+                });
+            }
+
+            return;
+        }
+    }
+
+    private persistLaunchConfig(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration, change: IMainClassOption): void {
+        if (!change) {
+            return;
+        }
+
+        const launchConfigurations: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("launch", folder);
+        const rawConfigs: vscode.DebugConfiguration[] = launchConfigurations.configurations;
+        const targetIndex: number = this.findDebugConfig(rawConfigs, config);
+        if (targetIndex >= 0) {
+            rawConfigs[targetIndex].mainClass = change.mainClass;
+            rawConfigs[targetIndex].projectName = change.projectName;
+
+            launchConfigurations.update("configurations", rawConfigs);
+        }
+    }
+
+    private findDebugConfig(rawConfigs: vscode.DebugConfiguration[], config: vscode.DebugConfiguration): number {
+        const candidates = [];
+        for (let i = 0; i < rawConfigs.length; i++) {
+            if (rawConfigs[i].name === config.name) {
+                candidates.push(i);
+            }
+        }
+
+        if (candidates.length > 1) {
+            for (const index of candidates) {
+                if (this.compareDebugConfig(rawConfigs[index], config)) {
+                    return index;
+                }
+            }
+
+            return candidates[0];
+        } else if (candidates.length === 1) {
+            return candidates[0];
+        }
+
+        return -1;
+    }
+
+    private compareDebugConfig(config1: object, config2: object): boolean {
+        for (const key of Object.keys(config1)) {
+            if (config1[key] !== config2[key]) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private async chooseMainClass(folder: vscode.Uri | undefined): Promise<IMainClassOption | undefined> {
+        const res = await resolveMainClass(folder);
         if (res.length === 0) {
             utility.showErrorMessageWithTroubleshooting({
                 message: "Cannot find a class with the main method.",
@@ -204,7 +296,13 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
             });
             return undefined;
         }
-        const pickItems = res.map((item) => {
+
+        return await this.showMainClassQuickPick(res, "Select main class<project name>");
+    }
+
+    private async showMainClassQuickPick(candidates: IMainClassOption[], quickPickHintMessage: string, alwaysShowQuickPickBox?: boolean):
+        Promise<IMainClassOption | undefined> {
+        const pickItems = candidates.map((item) => {
             let name = item.mainClass;
             let details = `main class: ${item.mainClass}`;
             if (item.projectName !== undefined) {
@@ -216,11 +314,10 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                 label: name,
                 item,
             };
-        }).sort((a, b): number => {
-            return a.label > b.label ? 1 : -1;
         });
-        const selection = pickItems.length > 1 ?
-            await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" })
+
+        const selection = alwaysShowQuickPickBox || pickItems.length > 1 ?
+            await vscode.window.showQuickPick(pickItems, { placeHolder: quickPickHintMessage })
             : pickItems[0];
         if (selection && selection.item) {
             return selection.item;
@@ -243,6 +340,11 @@ function resolveMainClass(workspaceUri: vscode.Uri): Promise<IMainClassOption[]>
         return <Promise<IMainClassOption[]>>commands.executeJavaLanguageServerCommand(commands.JAVA_RESOLVE_MAINCLASS, workspaceUri.toString());
     }
     return <Promise<IMainClassOption[]>>commands.executeJavaLanguageServerCommand(commands.JAVA_RESOLVE_MAINCLASS);
+}
+
+function validateLaunchConfig(workspaceUri: vscode.Uri, mainClass, projectName): Promise<ILaunchValidationResponse> {
+    return <Promise<ILaunchValidationResponse>> commands.executeJavaLanguageServerCommand(commands.JAVA_VALIDATE_LAUNCHCONFIG,
+        workspaceUri ? workspaceUri.toString() : undefined, mainClass, projectName);
 }
 
 async function updateDebugSettings() {
@@ -281,4 +383,15 @@ function convertLogLevel(commonLogLevel: string) {
 interface IMainClassOption {
     readonly projectName?: string;
     readonly mainClass: string;
+}
+
+interface IValidationResult {
+    readonly passed: boolean;
+    readonly message?: string;
+}
+
+interface ILaunchValidationResponse {
+    readonly mainClass: IValidationResult;
+    readonly projectName: IValidationResult;
+    readonly proposals?: IMainClassOption[];
 }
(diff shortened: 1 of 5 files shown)


src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return !Array.isArray(configItems) || !configItems.length;
     }
 
-    private async chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption | undefined> {
-        const res = await resolveMainClass(folder ? folder.uri : undefined);
+    private async quickFixLaunchConfig(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration):
```
yaohaizh: I don't think this function is to fix the launch config; fix is only part of the functionality after failed to choose mainclass. More like populate the launchConfig.

testforstephen: IMO, when the mainClass and projectName is empty, and we prompt users to choose one, this is also one kind of fix. And the difference is we don't persist them in launch.json.

yaohaizh: Fix means something is wrong and you want to correct it. For user configuration, the mainClass and projectName is correct and doesn't need fix them after validation.

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                errors.push(String(validationResponse.mainClass.message));
+            }
+
+            if (!validationResponse.projectName.passed) {
+                errors.push(String(validationResponse.projectName.message));
+            }
+
+            if (validationResponse.proposals && validationResponse.proposals.length) {
+                const answer = await utility.showErrorMessageWithTroubleshooting({
+                    message: errors.join("\n"),
```
yaohaizh: Use platform dependent separator

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+
+            return candidates[0];
+        } else if (candidates.length === 1) {
+            return candidates[0];
+        }
+
+        return -1;
+    }
+
+    private compareDebugConfig(config1: object, config2: object): boolean {
```
yaohaizh: The convention of semantic for compare result is -1, 0, 1 but not boolean value. Please keep this convention as possible. 

Use equals*** for boolean result.

src/configurationProvider.ts
```diff
@@ -282,3 +384,14 @@ interface IMainClassOption {
     readonly projectName?: string;
     readonly mainClass: string;
 }
+
+interface IValidationResult {
+    readonly passed: boolean;
+    readonly message?: string;
```
yaohaizh: message ==> errorMessage?

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+
+            if (!validationResponse.projectName.passed) {
+                errors.push(String(validationResponse.projectName.message));
+            }
+
+            if (validationResponse.proposals && validationResponse.proposals.length) {
+                const answer = await utility.showErrorMessageWithTroubleshooting({
+                    message: errors.join("\n"),
+                    type: Type.USAGEERROR,
+                }, "Quick Fix");
```
akaroml: 'Quick Fix' -> 'Fix'

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                if (answer === "Quick Fix") {
+                    const quickFix: IMainClassOption = await this.showMainClassQuickPick(validationResponse.proposals,
```
akaroml: What if user cancels and returns nothing?

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                    message: errors.join("\n"),
+                    type: Type.USAGEERROR,
+                });
+            }
+
+            return;
+        }
+    }
+
+    private persistLaunchConfig(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration, change: IMainClassOption): void {
```
akaroml: The function looks generic from its name and signature. But the data type of chagne is restricted to IMainClassOption which makes it only accept main class changes.
To workaround this, the signature should look like (folder, oldConfig, newConfig) to me, newConfig is with change applied. oldConfig is for us to look up the actual section in the file.

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        if (targetIndex >= 0) {
+            rawConfigs[targetIndex].mainClass = change.mainClass;
+            rawConfigs[targetIndex].projectName = change.projectName;
+
+            launchConfigurations.update("configurations", rawConfigs);
+        }
+    }
+
+    private findDebugConfig(rawConfigs: vscode.DebugConfiguration[], config: vscode.DebugConfiguration): number {
+        const candidates = [];
```
akaroml: lodash.isEqual does deep comparison.
_.findIndex(rawConfigs, _.isEqual) can get you the index.

src/configurationProvider.ts
```diff
@@ -216,11 +314,10 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                 label: name,
                 item,
             };
-        }).sort((a, b): number => {
-            return a.label > b.label ? 1 : -1;
         });
-        const selection = pickItems.length > 1 ?
-            await vscode.window.showQuickPick(pickItems, { placeHolder: "Select main class<project name>" })
+
+        const selection = alwaysShowQuickPickBox || pickItems.length > 1 ?
```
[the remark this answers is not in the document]
testforstephen: The flag is not about focus thing. Just to indicate whether to show the QuickPick box when only one available pick item.

In previous logic, when there is only one mainClass available, then pick it silently. But in the Fix prompt, even only one available mainClass, i want to show the QuickPick box to keep user consent about the ongoing fix replacement.

src/configurationProvider.ts
testforstephen: If the user chooses to cancel the fix QuickPick box, then the debugger exits.

src/configurationProvider.ts
```diff
@@ -386,8 +377,8 @@ interface IMainClassOption {
 }
 
 interface IValidationResult {
-    readonly passed: boolean;
-    readonly message?: string;
+    readonly isValid: boolean;
+    readonly errorMessage?: string;
```
akaroml: The message field can be neutral because there could message for both valid and invalid case.

src/configurationProvider.ts
```diff
@@ -1,7 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as _ from "lodash";
+import * as os from "os";
 import * as vscode from "vscode";
+
```
akaroml: Nice to have extra line to separate 3rd party dependencies and local dependencies.

src/configurationProvider.ts
testforstephen: Agree. Have reverted the field back to `message`.

src/configurationProvider.ts
```diff
@@ -203,8 +206,86 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        if (!validationResponse.mainClass.isValid) {
+            errors.push(String(validationResponse.mainClass.message));
+        }
+
+        if (!validationResponse.projectName.isValid) {
+            errors.push(String(validationResponse.projectName.message));
+        }
+
+        if (validationResponse.proposals && validationResponse.proposals.length) {
+            const answer = await utility.showErrorMessageWithTroubleshooting({
```
andxu: dup with line 257, you can keep the promise of utility.showErrorMessageWithTroubleshooting and await it if validationResponse.proposals have values.

testforstephen: There is a tiny difference between if/else conditions here. For the case of having proposals, the UI shows a `Fix` button, and the debug session will wait the user to click it. For the case of no having proposals, the UI just shows a `Learn More` button, no need to wait the user's click behavior, and the debug session will terminate directly.

src/configurationProvider.ts
testforstephen: after discussion, we think the neutral `message` name is more appropriate, because message could be for both valid and invalid case.

src/configurationProvider.ts
```diff
@@ -195,16 +194,115 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        const targetIndex: number = this.findDebugConfig(rawConfigs, config);
+        if (targetIndex >= 0) {
+            rawConfigs[targetIndex].mainClass = change.mainClass;
+            rawConfigs[targetIndex].projectName = change.projectName;
+
+            launchConfigurations.update("configurations", rawConfigs);
+        }
+    }
+
+    private findDebugConfig(rawConfigs: vscode.DebugConfiguration[], config: vscode.DebugConfiguration): number {
```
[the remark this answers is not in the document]
testforstephen: Use lodash utility to search the debug config instead.

src/configurationProvider.ts
testforstephen: this method is discarded.

src/configurationProvider.ts
```diff
@@ -126,15 +128,16 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
-                    const userSelection = await this.chooseMainClass(folder);
-                    if (!userSelection || !userSelection.mainClass) {
-                        // the error is handled inside chooseMainClass
-                        return;
-                    }
-                    config.mainClass = userSelection.mainClass;
-                    config.projectName = userSelection.projectName;
+
+                const mainClassOption = await this.resolveLaunchConfig(folder ? folder.uri : undefined, config);
+                if (!mainClassOption || !mainClassOption.mainClass) {
```
akaroml: Use try/catch block to handle the rejected/exception case.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return Object.keys(config).filter((key: string) => key !== "noDebug").length === 0;
     }
 
-    private async chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption | undefined> {
-        const res = await resolveMainClass(folder ? folder.uri : undefined);
+    private isEmptyArray(configItems: any): boolean {
```
akaroml: Use `_.isEmpty()` instead.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        return !Array.isArray(configItems) || !configItems.length;
+    }
+
+    private async resolveLaunchConfig(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration): Promise<IMainClassOption> {
```
akaroml: If the first parameter is optional, then we could write it as `folder?: vscode.Uri`. The question mark after the param name will assign `undefined` value when the param is not omitted. "default to undefined" is not necessariy by the nature of ts/js.

I suggest that you place `config` param as the first param because it's required. And folder to the second because it's optional.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        }
+
+        return {
+            mainClass: config.mainClass,
+            projectName: config.projectName,
+        };
+    }
+
+    private async fixMainClass(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration, validationResponse: ILaunchValidationResponse):
+        Promise<IMainClassOption | undefined> {
```
akaroml: A promise should be resolved with real results instead of undefined. In the case that we are not able to resolve some real results, it looks like an exception case.

src/configurationProvider.ts
```diff
@@ -225,17 +306,24 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
 
-        if (selection && selection.item) {
-            this.debugHistory.updateMRUTimestamp(selection.item);
-            return selection.item;
-        } else {
-            return undefined;
+    private async showMainClassQuickPick(pickItems: IMainClassQuickPickItem[], quickPickHintMessage: string, autoPick: boolean = true):
+        Promise<IMainClassOption | undefined> {
+        if (pickItems.length === 0) {
+            return;
```
akaroml: return nothing means resolving with undefined. This does not seem right. I would recommend through an exception because an empty candidate list is an error case.

src/configurationProvider.ts
```diff
@@ -225,17 +306,24 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
         }
+
+        const selected = (pickItems.length === 1 && autoPick) ?
+            pickItems[0] : await vscode.window.showQuickPick(pickItems, { placeHolder: quickPickHintMessage });
```
akaroml: What if user cancels by pressing Esc?

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+            return await this.fixMainClass(folder, config, validationResponse);
+        }
+
+        return {
+            mainClass: config.mainClass,
+            projectName: config.projectName,
+        };
+    }
+
+    private async fixMainClass(folder: vscode.Uri | undefined, config: vscode.DebugConfiguration, validationResponse: ILaunchValidationResponse):
```
akaroml: Same comment regarding optional params.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                message: errors.join(os.EOL),
+                type: Type.USAGEERROR,
+                anchor: anchor.FAILED_TO_RESOLVE_CLASSPATH,
+            });
+        }
+
+        return;
+    }
+
+    private async persistMainClassOption(folder: vscode.Uri | undefined, oldConfig: vscode.DebugConfiguration, change: IMainClassOption):
```
akaroml: Same comment regarding optional params and sequence.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        Promise<void> {
+        const newConfig: vscode.DebugConfiguration = _.cloneDeep(oldConfig);
+        newConfig.mainClass = change.mainClass;
+        newConfig.projectName = change.projectName;
+
+        return this.persistLaunchConfig(folder, oldConfig, newConfig);
+    }
+
+    private async persistLaunchConfig(folder: vscode.Uri | undefined, oldConfig: vscode.DebugConfiguration, newConfig: vscode.DebugConfiguration):
```
akaroml: Same comment regarding optional params and sequence.

src/configurationProvider.ts
```diff
@@ -213,8 +212,90 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+        const launchConfigurations: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("launch", folder);
+        const rawConfigs: vscode.DebugConfiguration[] = launchConfigurations.configurations;
+        const targetIndex: number = _.findIndex(rawConfigs, (config) => _.isEqual(config, oldConfig));
+        if (targetIndex >= 0) {
+            rawConfigs[targetIndex] = newConfig;
+            await launchConfigurations.update("configurations", rawConfigs);
+        }
+    }
+
+    private async promptMainClass(folder: vscode.Uri | undefined): Promise<IMainClassOption | undefined> {
```
akaroml: Same comment regarding optional params.

src/configurationProvider.ts
testforstephen: `selected = undefined` if user cancels by ESC.

src/configurationProvider.ts
testforstephen: Actually the pattern is inherited from VSCode API. See the original VSCode API defintion `public resolveDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration, token?: vscode.CancellationToken):`

testforstephen: It indicates the `folder` parameter is not optional. Just represent it could be `undefined` value when no workspace is opened.

src/configurationProvider.ts
testforstephen: when user cancels the fix dropdown, should return `undefined` instead of an exception. because it's user consent behavior.

src/configurationProvider.ts
testforstephen: will remove the empty check logic, because the check has been done in the caller side. No need to duplicate it.

Issue #360, opened 2018-08-10
"Learn More" link should point to the associated troubleshooting paragraph directly


testforstephen:
Currently when the user clicks "Learn More" button, it just opens the whole troubleshooting page. It's better to jump to the paragraph associated with the target error directly.

testforstephen:
closed with commit https://github.com/Microsoft/vscode-java-debug/commit/ae23fcb02e252b3ff2b5e2fab8023ea89c812e71

PR #371, opened 2018-08-21
"Learn More" link jumps to the associated troubleshooting paragraph directly

testforstephen:
The requirement is from issue https://github.com/Microsoft/vscode-java-debug/issues/360

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -70,7 +70,7 @@ You configure the incorrect main class name in `mainClass` of *launch.json*.
 
 ### Try:
 1. Check whether the class name specified in `mainClass` exists and is in the right form.
-2. If the problem persists, it's probably because the language server doesn't load your project correctly. Please reference *"Java Language Support extension fails to start"* paragraph for more troubleshooting info.
+2. If the problem persists, it's probably because the language server doesn't load your project correctly. Please reference the [language server troubleshooting](#try) paragraph for more troubleshooting info.
 
 ## The DEBUG CONSOLE throws ClassNotFoundException
 ### Reason:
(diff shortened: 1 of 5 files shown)


Troubleshooting.md
```diff
@@ -18,6 +18,11 @@ The debugger works with [Language Support for Java(TM) by Red Hat](https://marke
 
 ## Failed to resolve classpath:
 ### Reason:
+Below are the common failure reasons.
```
yaohaizh: In the document, it should avoid using `xyz` which is not meaningful. 

Should use meaningful name such as `mainClassName`, `projectName`

testforstephen: Nice suggestion. thanks.

src/utility.ts
```diff
@@ -43,26 +44,27 @@ export async function showErrorMessage(message: IMessage, ...items: string[]): P
...
 
 function openLink(url: string): void {
     opn(url);
 }
 
-function handleTroubleshooting(choice: string, message: string): string | undefined {
+function handleTroubleshooting(choice: string, message: string, anchor: string): string | undefined {
     if (choice === LEARN_MORE) {
-        openLink(TROUBLESHOOTING_LINK);
+        const link = anchor ? `${TROUBLESHOOTING_LINK}#${anchor}` : TROUBLESHOOTING_LINK;
```
andxu: no need for a `link` variable since it is simple and used only once

src/utility.ts
```diff
@@ -43,26 +44,27 @@ export async function showErrorMessage(message: IMessage, ...items: string[]): P
...
 export async function showWarningMessageWithTroubleshooting(message: IMessage, ...items: string[]): Promise<string | undefined> {
     const choice = await showWarningMessage(message, ...items, LEARN_MORE);
-    return handleTroubleshooting(choice, message.message);
+    return handleTroubleshooting(choice, message.message, message.anchor);
 }
 
 export async function showErrorMessageWithTroubleshooting(message: IMessage, ...items: string[]): Promise<string | undefined> {
     const choice = await showErrorMessage(message, ...items, LEARN_MORE);
-    return handleTroubleshooting(choice, message.message);
+    return handleTroubleshooting(choice, message.message, message.anchor);
```
andxu: rename IMessage/message  to  ITroubleShooting/troubleShoot?

src/utility.ts
testforstephen: fixed.

src/utility.ts
testforstephen: fixed.

Issue #350, opened 2018-08-02
Remember least recently debugged main class


Eskibear:
When I have multiple main classes in current workspace, after pressing F5 the debugger pops up a list of main class to let me choose. (sorted alphabetically)

When I need to re-launch to debug the same main class, every time I have to look for it again in the list. Is it possible to remember the LRU main class, and put it ahead of the list. It improves my experience.

testforstephen:
Another feedback is to put the mainClass from current active file in the top of the list too.

testforstephen:
This feature only applies to the scenario where `mainClass` and `projectName` is emtpy. (See the following config)
```
        {
            "type": "java",
            "name": "Debug (Launch)",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopOnEntry": false,
            "mainClass": "",
            "args": ""
        }
```

testforstephen:
PR https://github.com/Microsoft/vscode-java-debug/pull/372 merged. Close it.

PR #372, opened 2018-08-22
Put recently used main class in the top

testforstephen:

This pr will implement feature https://github.com/Microsoft/vscode-java-debug/issues/350
1. Put recently used main class in the top.
2. Put main class from active editor in the top. If it's not the most recently used one, then put it to the second top.

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
-
+import * as path from "path";
 import * as vscode from "vscode";
 
 import * as anchor from "./anchor";
@@ -10,6 +10,8 @@ import * as utility from "./utility";
 
 export class JavaDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
     private isUserSettingsDirty: boolean = true;
+    private debugHistory: RecentlyUsedHistory = new RecentlyUsedHistory();
+
     constructor() {
         vscode.workspace.onDidChangeConfiguration((event) => {
             if (vscode.debug.activeDebugSession) {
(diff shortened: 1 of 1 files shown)


testforstephen:
The UX is as following:
image

src/configurationProvider.ts
```diff
@@ -290,4 +356,29 @@ function convertLogLevel(commonLogLevel: string) {
 interface IMainClassOption {
     readonly projectName?: string;
     readonly mainClass: string;
```
yaohaizh: Order

The optional should behind required

src/configurationProvider.ts
```diff
@@ -290,4 +356,29 @@ function convertLogLevel(commonLogLevel: string) {
...
+    readonly filePath?: string;
+}
+
+interface IMainClassQuickPick extends vscode.QuickPickItem {
+    item: IMainClassOption;
```
yaohaizh: QuickPick? It should be ending with `item` to make the naming consistent.

src/configurationProvider.ts
```diff
@@ -290,4 +356,29 @@ function convertLogLevel(commonLogLevel: string) {
...
+}
+
+class RecentlyUsedHistory {
+    private cache: { [key: string]: number } = {};
```
yaohaizh: Use the Map for this this purpose?

src/configurationProvider.ts
```diff
@@ -290,4 +356,29 @@ function convertLogLevel(commonLogLevel: string) {
...
     readonly projectName?: string;
     readonly mainClass: string;
+    readonly filePath?: string;
+}
+
+interface IMainClassQuickPick extends vscode.QuickPickItem {
+    item: IMainClassOption;
+}
+
+class RecentlyUsedHistory {
```
yaohaizh: Should be MostRecentlyUsedCache 

MRU or LRU. Recently is not accuracy here.

src/configurationProvider.ts
testforstephen: This data structure has no relationship with Recently Used. Will refine this interface to an appropriate name.

src/configurationProvider.ts
```diff
@@ -213,30 +215,94 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+
+        if (!file || !currentActiveFile) {
+            return false;
+        }
+
+        if (path.relative(file, currentActiveFile) === "") {
+            return true;
+        }
+
+        return false;
```
andxu: can combine all the boolean operations like return file && currentActiveFile && path.relative(file, currentActiveFile) === "")

testforstephen: fixed.

src/configurationProvider.ts
```diff
@@ -288,6 +346,31 @@ function convertLogLevel(commonLogLevel: string) {
...
+
+    public updateMRUTimestamp(mainClassOption: IMainClassOption): void {
+        this.cache[this.getId(mainClassOption)] = Date.now();
+    }
+
+    public contains(mainClassOption: IMainClassOption): boolean {
+        return Boolean(this.cache[this.getId(mainClassOption)]);
+    }
+
+    private getId(mainClassOption: IMainClassOption): string {
```
yaohaizh: getKey to be consistent with the cache property name?

src/configurationProvider.ts
```diff
@@ -213,30 +215,86 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
         if (selection && selection.item) {
+            this.debugHistory.updateMRUTimestamp(selection.item);
             return selection.item;
         } else {
             return undefined;
         }
     }
+
+    private isOpenedInActiveEditor(file: string): boolean {
+        const activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
```
yaohaizh: shortcut logic if !file

testforstephen: In last commit, i have combined the multiple if logic in one shortcut statement. `file && currentActiveFile && path.relative(file, currentActiveFile) === ""`.  Is this comment for that case?

src/configurationProvider.ts
testforstephen: fixed.

Issue #376, opened 2018-08-24
Request type "undefined" is not supported


sallyur:
##### Environment
- Operating System: Windows and Mac
- JDK version: JDK 8
- Visual Studio Code version: 1.26.1
- Java extension version: 0.29.0
- Java Debugger extension version: 0.11.0
- Build: 297

##### Steps To Reproduce
1. Open file Main.java in folder 21.single-file
2. Press ctrl+F5 to start without debugging
##### Result
Error: Request type "undefined" is not supported. Only "launch" and "attach" are supported.

![image](https://user-images.githubusercontent.com/28860891/44569699-d0db1680-a7ad-11e8-86d1-6a336c696def.png)

Eskibear:
fixed. Could you verify build 299?

sallyur:
Test passed.

Vscode version: 1.26.1
Build: 299

Now the behavior is:

1. Pressing Ctrl+F5 for a single file it start without debugging  
2. Pressing Ctrl+F5 for a first-time-opened project, it will generate a launch.json file. Pressing Ctrl+F5 again it start without debugging

PR #377, opened 2018-08-24
Robustness: judge no launch.json exists.

Eskibear:
Pressing <kbd>Ctrl+F5</kbd> for a single file / first-time-opened project, the initial config to resolve is `{ noDebug: true }`. 
Current conditional doesn't apply. 

See more in  https://github.com/Microsoft/vscode-java-debug/issues/376#issuecomment-415677697

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -101,11 +101,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
              * If no launch.json exists in the current workspace folder
              * delegate to provideDebugConfigurations api to generate the initial launch.json configurations
              */
-            if (!config.request && folder !== undefined) {
+            if (this.isEmptyConfig(config) && folder !== undefined) {
                 return config;
             }
             // If it's the single file case that no workspace folder is opened, generate debug config in memory
-            if (!config.request && !folder) {
+            if (this.isEmptyConfig(config) && !folder) {
                 config.type = "java";
                 config.name = "Java Debug";
                 config.request = "launch";
@@ -201,6 +201,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
 
     private isEmptyArray(configItems: any): boolean {
         return !Array.isArray(configItems) || !configItems.length;
+    }
+
+    private isEmptyConfig(config: vscode.DebugConfiguration): boolean {
+        // filter out the auto filled field "noDebug"
+        return Object.keys(config).filter((key: string) => key !== "noDebug").length === 0;
     }
 
     private async chooseMainClass(folder: vscode.WorkspaceFolder | undefined): Promise<IMainClassOption | undefined> {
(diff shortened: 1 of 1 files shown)


src/configurationProvider.ts
```diff
@@ -101,11 +101,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
              * If no launch.json exists in the current workspace folder
              * delegate to provideDebugConfigurations api to generate the initial launch.json configurations
              */
-            if (Object.keys(config).length === 0 && folder !== undefined) {
+            if (!config.request && folder !== undefined) {
```
testforstephen: should write a common function to detect whether it's an empty config. Empty check should filter out the auto filled field `noDebug`.

src/configurationProvider.ts
```diff
@@ -203,6 +203,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return !Array.isArray(configItems) || !configItems.length;
     }
 
+    private isEmptyConfig(config: vscode.DebugConfiguration): boolean {
```
yaohaizh: Should be isValidConfig

src/configurationProvider.ts
```diff
@@ -203,6 +203,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return !Array.isArray(configItems) || !configItems.length;
     }
 
+    private isValidConfig(config: vscode.DebugConfiguration): boolean {
```
testforstephen: The intention is not for checking if config is valid or not, but find the case that launch.json is missing or empty. So isEmptyConfig is more appropriate.

Eskibear: > So isEmptyConfig is more appropriate.

After some consideration, I agree that `isValidConfig` might not be proper.  

Two reasons:
1. Confusion. `isValidConfig` implies the config contains all the information needed to launch a debug session, but actually it doesn't. 
2. For the name `isValidConfig`, it should have different value for following 2 configs, one is valid and the other invalid. However it doesn't, the function simply counts the keys and ignores "noDebug". 

```jsonc
// a valid config:
        {
            "type": "java",
            "name": "Debug (Launch)",
            "request": "launch", 
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopOnEntry": false,
            "mainClass": "",
            "args": ""
        }
// an invalid config:
       {
            "type": "java",
            "name": "Debug (Launch)",
            "request": "invalidValue",   // <------HERE
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopOnEntry": false,
            "mainClass": "",
            "args": ""
        }
```

For the name `isEmptyConfig`, yes we just regard `{noDebug: true}` as "empty", it's exactly why we need a helper function to customize the way we judge an empty config. Finally I'm ok with the name.
@yaohaizh are you ok with the name, or a better candidate if possible.

PR #384, opened 2018-08-29
Ignore auto-generated files in test projects.

Eskibear:
When developing the extension, the language server somehow is activated and recognizes the gradle projects. It auto generates some files. It's better to ignore them in case of being mistakenly checked in.

### DIFF
--- a/testprojects/12.gradle/.gitignore
+++ b/testprojects/12.gradle/.gitignore
@@ -1,0 +1,6 @@
+# Language server somehow recognizes build.gradle and compile this project.
+# Ignore following generated files to keep SCM clean.
+.gradle/
+.settings/
+.project
+.classpath


testprojects/12.gradle/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Language server somehow recognizes build.gradle and compile this project.
+# Ignore following generated files to keep SCM clean.
+.gradle/
+.settings/
+.project
```
testforstephen: some test project is eclipse project itself, how to handle the case?

Eskibear: Note that this file is under the test project 12, it only takes effect within the folder

andxu: please add .idea

Eskibear: I'm not going to do that in this PR. `.idea` should be added in the global .gitignore file (if needed). 
Actually I don't think we need to add that because this project is not supposed to be open in IntelliJ IDEA.

PR #404, opened 2018-09-06
add test plan for runtime class entries

andxu:


andxu:
dup as https://github.com/Microsoft/vscode-java-debug/pull/405

Issue #375, opened 2018-08-23
Ability to run java program in much simpler way


yaohaizh:
_From @patelkunal on August 22, 2018 12:40_

It would be great functionality to run java main class/program from visual code on click of annotation like Intellij (every main method annotated with nice run green icon). That makes it very easy to just run the java program/main method without any hassle.

Currently debugger-for-java plugin serves the work around by adding entry into launch.json and then select the class from drop-down list (finding it little difficult because I have ten/twenty main method/class inside my workspace). It would be awesome that I can invoke my java program based on context of the main editor with main method annotation. And since this plugin knows the classpath based on either maven/gradle builds, it would be easier to invoke the main method from the annotated icon. Again this is my wish to have it but it would improve developer experience up-to great extent.

Attaching reference images about requested feature
<img width="556" alt="screen shot 2018-08-22 at 6 06 15 pm" src="https://user-images.githubusercontent.com/2404315/44463697-34423880-a636-11e8-95bd-d9e5a449a4af.png">
<img width="498" alt="screen shot 2018-08-22 at 6 06 24 pm" src="https://user-images.githubusercontent.com/2404315/44463698-34423880-a636-11e8-88e5-a9fd55ccb66c.png">

I have also been using vscode for tutoring algos/cs-problem-solving where I am using have multi-modules build with many-many main classes (because its easy to explain the concept rather than worrying about best-practices at that moment). And when I press F5, it gives me long list of dropdown to select my main class which I want to run.

I have also explored code-runner plugin but it just dumb way of typing out command and its difficult to work with classpath and other modules settings. And again I have to maintain commands for all of my main class - why can't that main method can be identified by plugin and get nice annotation to make it easier to run.

##### Environment
- Operating System: MacOS
- JDK version: 1.8.0_172
- Visual Studio Code version: 1.26.1
- Java extension version: 0.29.0

Feel free to mark duplicate if this is already been handled (though couldn't find in issues here).

Thanks in advance,
/kp

_Copied from original issue: redhat-developer/vscode-java#615_

yaohaizh:
@testforstephen @Eskibear Please take a look at this issue. It should be related to your current work items.

yaohaizh:
_From @fbricon on August 23, 2018 2:43_

While I appreciate the idea of making it easier to run Java apps from the editor, this is out of scope of vscode-java. 
Please open a ticket in https://github.com/Microsoft/vscode-java-debug/issues

[...]

PR #410, opened 2018-09-06
Use codeLens to run Java program

testforstephen:

Feature request: https://github.com/Microsoft/vscode-java-debug/issues/375 

Please help review another PR on java side https://github.com/Microsoft/java-debug/pull/222 (an utility method to resolve main methods from a java file).

Currently pending issue for this feature is it doesn't support update codeLens dynamically when the user's modification is not saved. Only the file is saved, the codeLens provider will be notified, and refresh codeLens.

### DIFF
--- a/src/debugCodeLensProvider.ts
+++ b/src/debugCodeLensProvider.ts
@@ -54,12 +54,12 @@ class DebugCodeLensProvider implements vscode.CodeLensProvider {
     }
 }
 
-function runJavaProgram(mainClass: string, projectName: string, uri: vscode.Uri): void {
-    runCodeLens(mainClass, projectName, uri, true);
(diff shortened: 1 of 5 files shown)


testforstephen:
The ux for debug codeLens:
image

src/debugCodeLensProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+                    tooltip: "Debug Java Program",
+                    arguments: [ method.mainClass, method.projectName, document.uri ],
+                }),
+            ];
+        }));
+    }
+}
+
+function runJavaProgram(mainClass: string, projectName: string, uri: vscode.Uri): void {
+    runCodeLens(mainClass, projectName, uri, true);
```
andxu: add return

testforstephen: fixed.

Issue #389, opened 2018-08-29
Feature Proposal: Make vmArgs accept an array


PeterLaComb-Availity:
I would like to propose that vmArgs accept an Array as an alternative to the current String. Keep the string for compatibility reasons, but allow an array because some projects accept a large number of arguments that would be easier to look at if presented as an array.

andxu:
The PR is attached, to test it, add the following properties to launch.json of test case ArgsTest(in TestPlan.md)
```
"mainClass": "ArgsTest",
"args": [" \"bar"],
"vmArgs": ["-Xms64M",  "-Xmx128M",  "-Dfoo= \\bar"]
```
Add the following statement to get the properties:
```
  String foo = (String) System.getProperties().get("foo");
  List<String> proList = Arrays.asList(args);
```
the variables should be like this:
![image](https://user-images.githubusercontent.com/17075863/45271039-3893a000-b4d6-11e8-8bef-d97b13005148.png)

PR #414, opened 2018-09-10
Accept both string and string array in args and vmArgs.

andxu:


andxu:
https://github.com/Microsoft/vscode-java-debug/issues/389

src/configurationProvider.ts
```diff
@@ -202,6 +211,22 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
     }
 
+    /**
+     * Converts an array of arguments to a string as the args and vmArgs.
+     */
+    private concatArgs(args: any[]): string {
+        return _.join(_.map(args, (arg: any): string  => {
+            const str = String(arg);
+            // if it has quotes or spaces, use double quotes to wrap it
+            if (/['"\s]/.test(str)) {
```
yaohaizh: What if the user already quote the arguments by himself/herself?

andxu: I will quote it twice and use may notice this soon and remove his quotes.

Issue #403, opened 2018-09-06
Update the oct icon for active editor to file


testforstephen:
Currently we show a clock icon for active editor, 
![image](https://user-images.githubusercontent.com/14052197/45129978-3e794080-b1b9-11e8-8fea-08113e0df8b6.png)

It should be file icon 
![image](https://user-images.githubusercontent.com/14052197/45130030-7f715500-b1b9-11e8-841d-e014415faccf.png)

testforstephen:
Additional issues about the prompt wizard:
1. If the active editor Java file contains multiple main class, the current prompt wizard only display the first found main class with active editor flag. This is a bug, it should show all main classes from it with active editor flag.
2. If the first option is the most recently one, and also in active editor, then the prompt wizard should also contain active editor flag. Currently it just show the tip word `active editor`, but don't have the `file` icon. This is a bug. The fix is also showing the `file` icon.

PR #419, opened 2018-09-11
Change active editor oct icon to file-text instead of clock

testforstephen:

Fix issue https://github.com/Microsoft/vscode-java-debug/issues/403

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -382,7 +382,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
             if (pickItems[positionForActiveEditor].detail) {
                 pickItems[positionForActiveEditor].detail += `, active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
             } else {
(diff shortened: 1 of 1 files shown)


src/configurationProvider.ts
```diff
@@ -382,7 +382,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
             if (pickItems[positionForActiveEditor].detail) {
                 pickItems[positionForActiveEditor].detail += `, active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
             } else {
-                pickItems[positionForActiveEditor].detail = `$(clock) active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
+                pickItems[positionForActiveEditor].detail = `$(file-text) active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
```
Eskibear: I don't see any "file-text" in octicons. Is it [this one](https://octicons.github.com/icon/file/)?

testforstephen: vscode doesn't recognize `$(file)`.  Refer to vscode source code about the supported list.
https://github.com/Microsoft/vscode/blob/master/src/vs/base/browser/ui/octiconLabel/octicons/octicons.less#L94

PR #426, opened 2018-09-13
Fix the display order and icon issues for main class prompt wizard

testforstephen:

Fix the new found issues for bug https://github.com/Microsoft/vscode-java-debug/issues/403.

It contains the following changes:
- Fix one active editor containing multiple main class issue.
image
- Fix the missed icon for active editor.
image

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -351,59 +351,54 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         return selected && selected.item;
     }
 
-    private isOpenedInActiveEditor(file: string): boolean {
-        const activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
-        const currentActiveFile: string = activeEditor ? activeEditor.document.uri.fsPath : undefined;
-
-        return file && currentActiveFile && path.relative(file, currentActiveFile) === "";
-    }
-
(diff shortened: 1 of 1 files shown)


src/configurationProvider.ts
```diff
@@ -364,38 +364,47 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
 
-        if (positionForActiveEditor >= 0) {
-            if (pickItems[positionForActiveEditor].detail) {
-                pickItems[positionForActiveEditor].detail += `, active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
-            } else {
-                pickItems[positionForActiveEditor].detail =
-                `$(file-text) active editor (${path.basename(options[positionForActiveEditor].filePath)})`;
+            if (this.isOpenedInActiveEditor(pickItem.item.filePath)) {
+                if (adjustedDetail) {
+                    adjustedDetail += `, $(file-text) active editor (${path.basename(pickItem.item.filePath)})`;
```
Eskibear: I think following way is better for adding new "tag" in future. 
```ts
const adjustedDetails  = [];
if (...) adjustedDetails.push(...);
if (...) adjustedDetails.push(...);
....
pickItem.detail = adjustDetails.join(", ");
```

Issue #366, opened 2018-08-14
Always launching jre instead of JDK


andxu:
Steps:
 Open any project  and write "new String("");"  and step in the String constructor, 

The variable should be two:
1. original
2. this

but actual the parameter for String constructor is arg0 which indicating that the program is running at jre env,eg:
```
n:\jdk1.8.0_181\jre\bin\java -Xdebug -Xnoagent -Djava.compiler=NONE -Xrunjdwp:transport=dt_socket,address=localhost:9316,server=n,suspend=y -Dfile.encoding=UTF-8 -cp C:\work\vscode-java-debug\testprojects\8.nosource\bin;C:\work\vscode-java-debug\testprojects\8.nosource\lib\test1.jar;C:\work\vscode-java-debug\testprojects\8.nosource\lib\commons-io-2.5.jar NoSourceTest
```

testforstephen:
The Java debugger will use the API `System.getProperty("java.home")` to resolve java bin path, and use it to launch the debuggee. For windows, the value of `System.getProperty("java.home")`  is read from the registry `HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft`, and it's generally pointing to the JRE.  This is the reason why the debuggee is launched in JRE instead of JDK.

To fix the issue, we need read the java bin path from the user environment variable `JAVA_HOME`.

PR #428, opened 2018-09-14
Use Language Server consistent java home to launch debuggee

testforstephen:

It fixes the issue https://github.com/Microsoft/vscode-java-debug/issues/366.

This PR will fix the JDK instead of JRE issue, but not fix the parameter issue for JDK library, where the JDK method still shows `arg0, arg1, ...`. This is a known issue for JDK debugging. See the doc http://www.thejavageek.com/2016/04/03/debug-jdk-source-code/ and https://dev.zhourenjian.com/blog/2007/08/04/debug-into-rtjar.html

### DIFF
--- a/src/utility.ts
+++ b/src/utility.ts
@@ -1,7 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as expandHomeDir from "expand-home-dir";
+import * as findJavaHome from "find-java-home";
 import * as opn from "opn";
+import * as path from "path";
+import * as pathExists from "path-exists";
 import * as vscode from "vscode";
 import { logger, Type } from "./logger";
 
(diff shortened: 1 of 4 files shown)


testforstephen:
The peered Java side PR: https://github.com/Microsoft/java-debug/pull/225

src/utility.ts
```diff
@@ -109,3 +113,25 @@ export function formatErrorProperties(ex: any): IProperties {
 
     return properties;
 }
+
+export function checkJavaHome(): Promise<string> {
```
andxu: getJavaHome

src/utility.ts
```diff
@@ -109,3 +113,25 @@ export function formatErrorProperties(ex: any): IProperties {
...
+    return new Promise((resolve, reject) => {
+        let javaHome: string = readJavaConfig() || process.env.JDK_HOME || process.env.JAVA_HOME;
+        if (javaHome) {
+            javaHome = expandHomeDir(javaHome);
+            if (pathExists.sync(javaHome) && pathExists.sync(path.resolve(javaHome, "bin", `javac${EXE_SUFFIX}`))) {
+                return resolve(javaHome);
+            }
+        }
+
+        findJavaHome((err, home) => {
```
andxu: will findJavaHome expand the home dir like above expandHomeDir?

testforstephen:
@yaohaizh Thanks for the concern. I created an issue on vscode-java about exporting the resolve java home info. See the issue link https://github.com/redhat-developer/vscode-java/issues/639

src/utility.ts
testforstephen: nice catch.

PR #455, opened 2018-10-09
Dispose telemetry wrapper on deactivation.

Eskibear:
Dispose the inner "reporter".

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,7 +3,7 @@
 
 import * as path from "path";
 import * as vscode from "vscode";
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -67,6 +67,7 @@ function initializeExtension(operationId: string, context: vscode.ExtensionConte
 
 // this method is called when your extension is deactivated
 export function deactivate() {
+    disposeTelemetryWrapper();
```
testforstephen: await

PR #471, opened 2018-11-09
Support for an alternative jvm

RobertDeRose:
* Added new launch configuration property `jvm`
* Changed `updateDebugSettings` to optionally take in the config object
  - uses `config.jvm` if available instead for calling `getJavaHome`

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,11 @@
                 "description": "%java.debugger.launch.stopOnEntry.description%",
                 "default": true
               },
+              "javaHome": {
+                "type": "string",
+                "description": "Path to an alternative JVM to run on. For example, Java Language Server is running on JDK 8 64-bit, but your app only works on 32-bit",
+                "default": "^\"\\${env:JAVA_HOME}\""
+              },
               "console": {
                 "type": "string",
                 "enum": [
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -19,6 +19,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
     private isUserSettingsDirty: boolean = true;
     private debugHistory: MostRecentlyUsedHistory = new MostRecentlyUsedHistory();
     private resolver: VariableResolver;
+
     constructor() {
         this.resolver = new VariableResolver();
         vscode.workspace.onDidChangeConfiguration((event) => {
@@ -95,7 +96,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
     private resolveVariables(folder: vscode.WorkspaceFolder, config: vscode.DebugConfiguration): void {
         // all the properties whose values are string or array of string
         const keys = ["mainClass", "args", "vmArgs", "modulePaths", "classPaths", "projectName",
-            "env", "sourcePaths", "encoding", "cwd", "hostName"];
+            "env", "sourcePaths", "encoding", "cwd", "hostName", "javaHome"];
         if (!config) {
             return;
         }
@@ -131,7 +132,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         try {
             if (this.isUserSettingsDirty) {
                 this.isUserSettingsDirty = false;
-                await updateDebugSettings();
+                await updateDebugSettings(config);
             }
 
             /**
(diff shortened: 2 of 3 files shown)


RobertDeRose:
inkedshow_vscode-java-debug_running_alternative_jvm_li

RobertDeRose:
ping @Eskibear What do you think about this feature? Anything I missed that you can see? I personally need it for VSCode to even be a possible editor for my work.

yaohaizh:
@testforstephen @andxu Would you please help evaluate this PR?

Eskibear:
> For example, Java Language Server is running on JDK 8 64-bit, but your app only works on 32-bit

As a dependent, java-debugger uses the same JAVA_HOME with language server for consistency. Personally, I don't understand the reason why you have to run language server on 64-bit and debug the app on 32-bit. Can you share the motivation? And can you simply change settings `java.home` to your 32-bit JDK?

RobertDeRose:
@Eskibear The motivation is that my project (inherited, new job, old code base) is a rather large (106MB of source code) monolithic code base.  I need Java 64-bit for the JLS (Java Language Server) because it needs well over 4 GB of RAM just to be usable, which is not possible when running on the 32-bit JVM. 

**Side note**, *the use of `${env:JAVA_HOME}` as the default is because the JLS defaults to using ${JDK_HOME} before ${JAVA_HOME} so this was an easy way to set a different JDK for debugging than the one the JLS uses.* 

That's the primary motivation.  The other, this just makes it easier to test your code against alternative JVM installs without need to constantly change and restart the VSCode (Which for me, is a slow process that can take up to 5 minutes to become stable enough to start working). So if for example I wanted to test against the newly announced Amazon Corretto JVM, or the Redhat compiled JVM, or maybe GraalVM (no idea how to use that yet, just throwing out examples).

The only issue I am having with my PR is that when I change the `jvm` setting I introduced, it doesn't seem to update the value it sends to the JLS, it works perfectly on initialization of VSCode and the extension, so I assume I am missing some change event to the `launch.json` that I either need to hook into or add so that the config object is re-created.

RobertDeRose:
@Eskibear Any thoughts? Would love to see this merged, but I would love even more feedback on where to add the needed change to ensure that if the config file is changed, the value I pass in via `jvm` is updated as well.

src/configurationProvider.ts
```diff
@@ -128,7 +129,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         try {
             if (this.isUserSettingsDirty) {
```
andxu: Could you please help to change the older mechanism of detecting setting dirty of using vscode.workspace.onDidChangeConfiguration and  affectsConfiguration(see debugCodeLensProvider.ts)
instead of isUserSettingsDirty

DebugCodeLensContainer

package.json
```diff
@@ -131,6 +131,11 @@
                 "description": "Automatically pause the program after launching.",
                 "default": true
               },
+              "jvm": {
```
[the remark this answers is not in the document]
RobertDeRose: Ok

src/configurationProvider.ts
RobertDeRose: Ok, I’ll take a look

andxu: Please also update the branch too.

zsalab:
This feature would be really useful, as I wish to run my Java app with DCEVM instead of standard OpenJDK VM. There is no need for advanced hotswap capabilities for running the language server but I wish to run my app under development with DCEVM advanced hotswap capability.
https://github.com/TravaOpenJDK/trava-jdk-11-dcevm

RobertDeRose:
> This feature would be really useful, as I wish to run my Java app with DCEVM instead of standard OpenJDK VM. There is no need for advanced hotswap capabilities for running the language server but I wish to run my app under development with DCEVM advanced hotswap capability.
> https://github.com/TravaOpenJDK/trava-jdk-11-dcevm

I've given up getting this in.. the request from the maintainers are far outside the scope of the purpose of this PR. I was never able to find time to comply. 

I've now switch to Python as my day work language so if anyone would like to fork this PR and get it up to the maintainers standards, I wish them well

RobertDeRose:
This was super useful, sad guys repo owners couldn't accept it for what it was

Issue #476, opened 2018-11-13
Simplify generated launch.json


isidorn:
Hey,

VSCode dev here. This milestone I am looking into simplifing generated `launch.json` for various extensions Microsoft/vscode#62851

The `launch.json` that Java generates is attached at the end. This is not too complex for the avarage user, however we can still improve. I suggest to simplify it the following way:

- [ ] `DebugConfigurationProvider` should use the `quickPick` to ask the user if he would like to launch or to attach. After that it should ask what is the `mainClass` or what is the `port`. Optimally it should be able to detect those things on its own, but if not possible use `quickPick` to ask the user.
- [ ] Remove default fields which are not necessery like `console`, `args`, `stopOnEntry`, `hostName` and `cwd`
- [ ] You are already contributing `configuraitonSnippets` which is great. However these snippets also need to be simplified. I suggest to remove all attributes that have the default value. Example: `stopAtEntry`, `cwd`, `args`

If you agree with the suggestions I am making here I am also here to help with any potential questions you might have. The changes should not require a lot of work but will simplify the flow a lot imho. It should be much less complex and not too much like a wizard experience

```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "java",
            "name": "Debug (Launch)",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "console": "internalConsole",
            "stopOnEntry": false,
            "mainClass": "",
            "args": ""
        },
        {
            "type": "java",
            "name": "Debug (Attach)",
            "request": "attach",
            "hostName": "localhost",
            "port": "<debug port of remote debuggee>"
        }
    ]
}
```

isidorn:
@testforstephen Are there any plans to address this?
Do you agree with my suggesstions here?
If you do not have cycles to address this but would like to see it done, I can try to get some users to provide a PR which tackles this.

Thanks a lot

testforstephen:
@isidorn Is there any strong reason to shorten the launch.json? Because we found the user will be pretty hard to discover the hiding launch options. 

Here is my thoughts about the removable options:
[...]

Issue #431, opened 2018-09-18
Introduce shortcut to launch program from active editor


testforstephen:
Open a new issue to track the additional discussion for the previous issue https://github.com/Microsoft/vscode-java-debug/issues/375.

At issue #375, it has provided codeLens to run/debug the program more conveniently. The discussion is to provide a single keyboard shortcut to launch program from active editor.

testforstephen:
In launch.json, add the following config. 
```
{
     "type": "java",
     "name": "Debug (Launch) - Current File",
     "request": "launch",
     "mainClass": "${file}"
}
```
And let Debug Viewlet select `Debug (Launch) - Current File` as default config.
![image](https://user-images.githubusercontent.com/14052197/49841997-d314bf80-fdf4-11e8-88c5-e6c5aeac5560.png)

Then F5, the debugger will auto launch current active Java file.

PR #481, opened 2018-11-29
Refine the launch json

testforstephen:

It's for issue `Simplify generated launch.json` #476 and `Introduce shortcut to launch program from active editor` #431.

3 refinements:
- Don't generate `cwd` and `stopOnEntry` by default, but keep `hostName` because it's easy to understand and no burden on learning curve.
- Provide shortcut to launch program from active editor. The user could specify `mainClass: "${file}"` for that, `F5` will auto detect main class from current file.
- Refine the hint message when prompting for main class.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
               },
               "mainClass": {
                 "type": "string",
-                "description": "The main class of the program (fully qualified name, e.g. [mymodule/]com.xyz.MainClass).",
+                "description": "The fully qualified class name (e.g. [java module name/]com.xyz.MainApp) or the java file path of the program entry.",
                 "default": ""
               },
               "args": {
@@ -118,7 +118,7 @@
               },
               "cwd": {
                 "type": "string",
-                "description": "The working directory of the program.",
+                "description": "The working directory of the program. Defaults to current workspace root.",
                 "default": "${workspaceFolder}"
               },
               "env": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -118,7 +118,7 @@
               },
               "cwd": {
                 "type": "string",
-                "description": "The working directory of the program.",
+                "description": "The working directory of the program. Defaults to current workspace root.",
```
andxu: the current

package.json
```diff
@@ -290,23 +290,31 @@
...
             }
           },
+          {
+            "label": "Java: Launch Program in Current File",
+            "description": "Add a new configuration for launching current java file.",
+            "body": {
+              "type": "java",
+              "name": "Debug (Launch) - Current File",
+              "request": "launch",
+              "console": "internalConsole",
```
andxu: I think we can hide the console and args, see the sample generated by node.js
```
{
            "type": "node",
            "request": "launch",
            "name": "Launch Program",
            "program": "${workspaceFolder}/app.js"
        }
```

yaohaizh: I agree with Andy here. Make our configuration consist with other language and common VSCode settings. The console/args should be familiar to the customer if they are using VSCode.

testforstephen: Removed `console`, `args`, and `attach` config from the default launch.json, and use the Config Snippet to improve the discovery.

testforstephen:
Removed `console`, `args`, and `attach` config from the default launch.json, and use the Config Snippet to improve the discovery.

Issue #462, opened 2018-10-19
[Feature]add a new option for disabling the build and use custom build in prelaunch task


andxu:
[provide a description of the issue]

##### Environment
- Operating System:
- JDK version:
- Visual Studio Code version:
- Java extension version:
- Java Debugger extension version:

##### Steps To Reproduce
1. [step 1]
2. [step 2]

[attach a sample project reproducing the error]
[attach logs](https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#enable-logging)

JamesBewley:
+1

testforstephen:
Add a new user setting `java.debug.settings.forceBuildBeforeLaunch` to control whether to build the workspace before launch/debug.

PR #485, opened 2018-11-30
Add an user setting to control whether to build the workspace before debugging

testforstephen:

add a new option for disabling the build and use custom build in prelaunch task #462


package.json
```diff
@@ -394,6 +394,11 @@
           "type": "boolean",
           "description": "%java.debugger.configuration.enableRunDebugCodeLens.description%",
           "default": true
+        },
+        "java.debug.settings.forceBuildBeforeLaunch": {
+          "type": "boolean",
+          "description": "%java.debugger.configuration.forceBuildBeforeLaunch%",
```
yaohaizh: Just curiously about the nls file. Could we group them together under some specified folder but not under root. It will be a disaster if we support more languages in future.

PR #514, opened 2018-12-13
Add to more categories.

Eskibear:


package.json
```diff
@@ -26,7 +26,9 @@
   },
   "homepage": "https://github.com/Microsoft/vscode-java-debug/blob/master/README.md",
   "categories": [
-    "Debuggers"
+    "Debuggers",
```
yaohaizh: Could add categories `Java`?

Eskibear: AFAIK, the value is a enum. No such category called Java.

PR #617, opened 2019-07-23
Remove hard dependency of redhat.java

Eskibear:
Signed-off-by: Yan Zhang <yanzh@microsoft.com>

This PR removes hard dependency of extension `redhat.java`, in order not to unneccessarily activate it. So `redhat.java` will not be activated when debugging other language, esp. when user don't have any java related files in workspace.

Debugger depends on Java extension on:
- pass params to execute delegate Java workspace command.
- execute other VS Code commands registered in extension `redhat.java`.

Thus I check visibility (installed/enabled status) of extension `redhat.java`. If it's absent, prompt to guide users to install it.


Eskibear:
If Java Ext is not installed, below is what the prompt box looks like:
Screen Shot 2019-07-23 at 11 01 00 AM

Eskibear:
debugger-install-java-ext

src/configurationProvider.ts
```diff
@@ -80,10 +81,13 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                         };
                     });
                     resolve([defaultLaunchConfig, ...launchConfigs]);
-                }, (ex) => {
+                } catch (ex) {
+                    if (ex instanceof utility.JavaExtensionNotActivatedError) {
+                        utility.guideToInstallJavaExtension();
```
testforstephen: The check logic may work for the scenario where Java debugger is activated mistakenly. But we also need take care of the normal Java debugging scenario. For example, when Java language server is activating, click F5, the user may receive an error message to guide to install Language Support for Java.

Eskibear: I should only pop up error message and guide user to install Java extension **ONLY WHEN** redhat.java is not installed. That is, `vscode.extensions.getExtension(JAVA_EXTENSION_ID)` returns the extension instance.

Issue #622, opened 2019-08-07
Non-ASCII char support for Windows terminals


Eskibear:
When launching in integrated/external terminal, the debugger passes commands to vscode, and vscode assembles and sends it to the terminal. But if there's non-ascii char in code/path, the java process will have unexpected behavior, because of bad Unicode support in Windows. 

Either VS Code or the extension should do something to cover these cases. 

## Example
Folder structure:
```bash
中文目录
├── Hello.class
└── Hello.java
0 directories, 2 files
```

Content of file `中文目录/Hello.java`:
```java
public class Hello {
  public static void main(String[] args) {
    System.out.println("中文");
  }
}
```

### a) on Mac OSX
Works as expected
```bash
$> java -cp 中文目录 Hello
中文
```
#### Non-ASCII char in classpath
Classpath is not correctly recognized by Java, if it contains non-ascii char.
```cmd
D:\Test>java -cp 中文目录 Hello
Error: Could not find or load main class Hello
```

#### Non-ASCII char in stdout
cd into the folder, the classpath is correctly recognized now, but the terminal fails to print non-ascii chars. 
```cmd
D:\Test>cd 中文目录

D:\Test\中文目录>java -cp . Hello
ä??æ??
```
To display non-ascii char in Windows terminals, you have to modify the `code page`, see https://ss64.com/nt/chcp.html

Workaround: set code page to 65001 (utf-8)
```cmd
D:\Test\中文目录>chcp 65001
Active code page: 65001

D:\Test\中文目录>java -cp . Hello
中文
```

Eskibear:
Currently in v0.21.0 we use a `launcher.bat` to set code page to 65001 before launching the java process. 
But it's reported to have issues:
* If IT Policy forbid to run .bat files, debugging is totally blocked. #646 
* It doesn't work well with _Git Bash for Windows_ for some users. #642 
* For MacOS, commands sent to integrated terminal sometimes is incomplete, thus debuggee process is not started, leading to debugger timeout. #614 #651

Eskibear:
> Non-ASCII char in classpath
> Classpath is not correctly recognized by Java, if it contains non-ascii char.
> ```
> D:\Test>java -cp 中文目录 Hello
> Error: Could not find or load main class Hello
> ```

Tried in 1.51.1, above issue has been fixed, but non-ascii stdout is still not working well. see below.


[...]

Issue #618, opened 2019-07-31
Unable to print Chinese characters correctly


EricZhangzf:
##### Environment
- Operating System: Windows10
- JDK version:1.8.0_211
- Visual Studio Code version:1.36.1
- Java extension version:0.47.0
- Java Debugger extension version:0.21.0-RC

##### Steps To Reproduce
1. Open a java project with Chinese characters.
2. Click F5 to run it.

##### Current Result
![image](https://user-images.githubusercontent.com/44559777/62194256-18015c80-b3ac-11e9-8e2a-5dcfea0f8798.png)

##### Expected Result
Result of Chinese characters.

Eskibear:
Try enabling system-wide UTF-8  support.
![image](https://user-images.githubusercontent.com/2351748/62259647-a116a300-b441-11e9-879c-bfe77ea75082.png)

EricZhangzf:
Fixed by check Beta: Use Unicode UTF-8 for worldwide language support.

Eskibear:
Ideally we should not require users to modify system-wide settings. Re-open to see if we have any alternative solutions.

Eskibear:
please keep the system-wide local unchanged, and verify the latest private build, see if the issue is gone.

EricZhangzf:
Test pass on 0.20.0

PR #625, opened 2019-08-08
Use a bat launcher to set page code

Eskibear:
See: #622 #618 
Require: https://github.com/microsoft/java-debug/pull/288

When runing in Windows terminal, use a launcher to set page code to UTF-8.

### DIFF
--- a/scripts/launcher.bat
+++ b/scripts/launcher.bat
@@ -1,0 +1,7 @@
+@echo off
+
+REM Change code page to UTF-8 for better compatibility.
+@chcp.com 65001 > NUL 
+
+REM Execute real command passed by args
+%*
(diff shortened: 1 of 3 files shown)


scripts/launcher.bat
```diff
@@ -0,0 +1,7 @@
+@echo off
+
+REM Change code page to UTF-8 for better compatibility.
+@chcp.com 65001 > NUL
```
testforstephen: why not to set the corresponding page according to the file.encoding parameter?

Eskibear: Always use UTF-8 for the moment (same behavior with debug console). Consider to improve only if there's feature request for it.

PR #630, opened 2019-08-13
User face change: say "Do not show again" instead of "Not show again".

pi1024e:
Typo fix because simply saying "Not show again" is weird to the user.


testforstephen:
thanks for the contribution.

PR #665, opened 2019-09-20
Don't support HCR in Run mode

testforstephen:

In Run mode, click ⚡ in debug toolbar, actually HCR will do nothing because we don't support hot replace in running mode.

An improvement is to show a message about the limitation.
image


Eskibear:
is it possible to hide the  ⚡ button if it's noDebug mode?

testforstephen:
it's registered via package.json statically. I only found `inDebugMode` when clause, but it doesn't provide signal about whether it's run or debug.

PR #679, opened 2019-09-30
Improve the system error

testforstephen:


src/extension.ts
```diff
@@ -178,10 +181,32 @@ async function runJavaFile(uri: vscode.Uri, noDebug: boolean) {
...
+        if (alreadyActivated) {
+            vscode.window.showErrorMessage(String((ex && ex.message) || ex));
+            return;
+        }
+
+        throw ex;
+    }
+
+    const activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
+    if (!uri && activeEditor && _.endsWith(path.basename(activeEditor.document.fileName), ".java")) {
```
jdneo: How about using `document.languageId`?

testforstephen: we don't support run a library class yet, e.g. "Main.class"

jdneo: I see.

PR #688, opened 2019-10-12
Prompt to add the folder to source path if the running file isn't on classpath

### LINKED ISSUE #470: Disable code lens for the java file not in the project's classpath
image

Hello.java is not in the maven source path `src/main/java`, so jdt will not persist the corresponding .class file in the `target` directory. When the user clicks the Run/Debug code lens, it will pop up `Error: Could not find or load main class Hello`. This sometimes will confuse the user, it's better that the debugger doesn't show the code lens.

testforstephen:

It requires https://github.com/microsoft/java-debug/pull/302
Close #470

When the Java file isn't on the source path of any project, JDT doesn't build it to .class. Run it via codelens or context menu, the runtime will throw class not found error. One improvement is to provide a fix option to ask user to add the parent folder to Java source path.

### DIFF
--- a/src/debugCodeLensProvider.ts
+++ b/src/debugCodeLensProvider.ts
@@ -169,8 +169,8 @@ async function constructDebugConfig(mainClass: string, projectName: string, work
 export async function startDebugging(mainClass: string, projectName: string, uri: vscode.Uri, noDebug: boolean): Promise<boolean> {
     const workspaceFolder: vscode.WorkspaceFolder = vscode.workspace.getWorkspaceFolder(uri);
     const workspaceUri: vscode.Uri = workspaceFolder ? workspaceFolder.uri : undefined;
-
-    if (workspaceUri && !(await isOnClasspath(uri.toString())) && !(await addToClasspath(uri))) {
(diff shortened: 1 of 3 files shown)


src/debugCodeLensProvider.ts
```diff
@@ -169,9 +170,35 @@ export async function startDebugging(mainClass: string, projectName: string, uri
...
+    if (!parentFsPath) {
+        return true;
+    }
+
+    const parentUri = vscode.Uri.file(parentFsPath);
+    let parentPath = vscode.workspace.asRelativePath(parentUri, true);
+    if (parentPath === parentUri.fsPath) {
+        parentPath = path.basename(parentFsPath);
+    }
+    const ans = await vscode.window.showWarningMessage(`The file ${fileName} isn't on the classpath, the runtime may throw class not found error. `
```
Eskibear: "in" or "on" the classpath?

testforstephen: JDT api called it `isOnClasspath`, i would to keep it consistent.

PR #733, opened 2019-12-20
Generate the launch configuration in memory if no launch.json exists

### LINKED ISSUE #724: Unhappy "Run Java" experience
As filled by @jrieken here https://github.com/microsoft/vscode/issues/86086

* open java-12 dev container
* rename .vscode folder to .vscode-off
* open debug viewlet, select "Run with Java"
* 🐛 no run... there is a progress message about generating configuration and a launch.json file is shown but that is all...

My comment: This is up to the Java extension, they can handle the No Workspace Run however the extension wants. I suggest that the Java extension handles run more gracefully and if the `launch.json` can get generated the program could be run as well. as part of the experience.

@testforstephen let me know what you think

testforstephen:

Fix #724

- No launch.json in the current workspace.
Triggering run or debug from the entries, such as **F5**, **Code Lens** and **Context Menu**, will generate a java configuration in memory and run the program directly.

- launch.json already exists in the current workspace.
Trigger from **Code Lens** or **Context Menu** will also persist the configuration into the launch.json.

### DIFF
--- a/src/debugCodeLensProvider.ts
+++ b/src/debugCodeLensProvider.ts
@@ -173,7 +173,7 @@ async function launchJsonExists(workspace: vscode.Uri): Promise<boolean> {
 
     const workspaceFolder = vscode.workspace.getWorkspaceFolder(workspace);
     const results: vscode.Uri[] = await vscode.workspace.findFiles(".vscode/launch.json");
-    return !!results.find((launchJson => vscode.workspace.getWorkspaceFolder(launchJson) == workspaceFolder));
+    return !!results.find((launchJson) => vscode.workspace.getWorkspaceFolder(launchJson) === workspaceFolder);
 }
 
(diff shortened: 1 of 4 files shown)


src/debugCodeLensProvider.ts
```diff
@@ -166,6 +166,11 @@ async function constructDebugConfig(mainClass: string, projectName: string, work
     return _.cloneDeep(debugConfig);
 }
 
+async function launchJsonExists(): Promise<boolean> {
+    const results: vscode.Uri[] = await vscode.workspace.findFiles(".vscode/launch.json", null, 1);
```
jdneo: Will this work in multi-root? How about `**/.vscode/launch.json`?

testforstephen: VS Code will search from each workspace root. This path is the path with the workspace root.

`**/.vscode/launch.json` will match `myProject/.settings/.vscode/launch.json` file, actually vscode will not take it as a launch.json.

PR #790, opened 2020-04-09
Show a warning message about the Unsupported JDK error

### LINKED ISSUE #789: Unsupported JDK may throw UnsupportedClassVersionError
The JDK release follows a strict 6 month cadence, when the user tries to run the standalone Java files with the latest JDK, sometimes they may meet the error UnsupportedClassVersionError. The reason is the extension may not catch up the support for the latest release yet.

Here is a typical error message. 
image

`57` is targeting to java 13, and `57.65535` is for java 13 preview.  similarly `58` is for java 14. Usually high versions are backward compatible with older versions. JVM 14 can run java 13 (**57**), but it doesn't support the old preview version (**57.65535**). The [openjdk](https://openjdk.java.net/jeps/12) website has claimed the reason that it would be costly for JDK 14 to support preview features from JDK 13 which were changed or dropped in response to feedback.

Reproduce steps:
- Installed Language Support for Java extension 0.58.0. (Java extension starts to support JDK 14 since 0.59.0).
- Install JDK 14 and set it to "java.home" user settings.
- Run standalone java files in VS Code, the output will throw the UnsupportedClassVersionError.

testforstephen:

It requires https://github.com/microsoft/java-debug/pull/326

Fixes #789

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -56,10 +56,14 @@ Below is a typical error message.
 
 ![image](https://user-images.githubusercontent.com/14052197/78854443-ed47c780-7a53-11ea-8317-d8b097dfba99.png)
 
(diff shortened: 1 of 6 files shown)


testforstephen:
image

PR #805, opened 2020-05-06
drop error event for some environment

testforstephen:

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -28,11 +28,7 @@ class Logger implements vscode.Disposable {
                 aiKey: extensionPackage.aiKey,
             };
             if (packageInfo.aiKey) {
-                this.reporter = new TelemetryReporter(packageInfo.name, packageInfo.version, packageInfo.aiKey);
-                if (firstParty) {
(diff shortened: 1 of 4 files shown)


src/logger.ts
```diff
@@ -29,6 +29,10 @@ class Logger implements vscode.Disposable {
             };
             if (packageInfo.aiKey) {
                 this.reporter = new TelemetryReporter(packageInfo.name, packageInfo.version, packageInfo.aiKey);
+                if (firstParty) {
+                    // @ts-ignore
+                    this.reporter.firstParty = firstParty;
```
Eskibear, line 34: new constructor in latest `vscode-extension-telemetry`, don't have to do such trick.

Eskibear: https://github.com/microsoft/vscode-extension-telemetry/commit/2fd787af0f317712cf982dfa5e00d38b5ef12f37#diff-90dddd74ec13ca1b38d1ae386ebc56f6R28-R30

testforstephen: nice catch!

PR #828, opened 2020-06-08
Auto enable -XX:+ShowCodeDetailsInExceptionMessages for java 14 apps

### LINKED ISSUE #797: Automatically add -XX:+ShowCodeDetailsInExceptionMessages to launch configurations
When a program is launched with Java 14, it might be useful to automatically add `-XX:+ShowCodeDetailsInExceptionMessages` to the vmArgs, so that NPEs provide better troubleshooting informations. See https://www.baeldung.com/java-14-nullpointerexception for more details.

The flag should be added dynamically during launch, if not already present in the launch.json's "vmArgs" option. Better not generate the launch.json with the flag directly in it, as things will go wrong if the project's JDK version changes:
> Unrecognized VM option 'ShowCodeDetailsInExceptionMessages'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

This enhanced behavior could be disabled with a new setting.

testforstephen:

Fix #797

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -11,7 +11,7 @@ import * as anchor from "./anchor";
 import { buildWorkspace } from "./build";
 import * as commands from "./commands";
 import * as lsPlugin from "./languageServerPlugin";
-import { detectLaunchCommandStyle, validateRuntime } from "./launchCommand";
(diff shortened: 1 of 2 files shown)


testforstephen:
Sample:
```java
System.out.println(person.getProfile().getEmail());
```

Before:
image

After:
image

PR #829, opened 2020-06-12
Skip the specified classes on exception breakpoint or stepping

### LINKED ISSUE #628: Support step "Just My Code"
Provide a feature to only step my files from my own project, and skip the library files. It's like 'Just my code' feature in Visual Studio.

### LINKED ISSUE #756: Break on exception in Java for “just my code”
##### Environment
- Operating System: Win10
- JDK version: 
- Visual Studio Code version: 1.41.1
- Java extension version: 0.55.1
- Java Debugger extension version: 0.24.0

##### Question
How can I configure VS Code to only break on an exception in "just my code"? I've messed around w/ the Uncaught Exceptions & Caught Exceptions but they are not scoped to my code.

Something like [Visual Studio -> Enable or disable Just My Code](https://docs.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2019#BKMK_Enable_or_disable_Just_My_Code).

testforstephen:

It requires https://github.com/microsoft/java-debug/pull/334

Close #756
Close #628


testforstephen:
When you just want to break on exception for "Just My Code", you can 
```json
    "java.debug.settings.exceptionBreakpoint.skipClasses": [
        "$JDK", // Skip the JDK classes.
        "$Libraries" // Skip the classes from the application libraries.
    ],
```

When you just want to step into "Just My Code", then you can 
```json
    "java.debug.settings.stepping.skipClasses": [
        "$JDK", // Skip the JDK classes.
        "$Libraries" // Skip the classes from the application libraries.
    ]
```

Also, you could add a specific class name expression, e.g. java.*, *.Foo to skip the classes.

src/classFilter.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function substituteFilterVariables(skipClasses: string[]): Promise<any> {
+    if (!skipClasses) {
+        return [];
+    }
+
+    try {
+        // Preprocess skipClasses configurations.
+        if (Array.isArray(skipClasses)) {
+            const index = skipClasses.findIndex((filter) => filter === "$JDK" || filter === "$Libraries");
```
jdneo, line 29: How about directly get the boolean value: `const hasReservedName = skipClasses.some((filter) => filter === "$JDK" || filter === "$Libraries");`

Eskibear:
Here at a user's perspective, I don't see the meaning of $JDK and $Libraries.
- $JDK, does it mean java.* and javax.*, or even including javafx.*?  Then why there's an explicit `java.lang.classloader` in default value?
- $Library. This is vague, it might be better to add more description?

testforstephen:
> Here at a user's perspective, I don't see the meaning of $JDK and $Libraries.
> 
> * $JDK, does it mean java.* and javax._, or even including javafx._?  Then why there's an explicit `java.lang.classloader` in default value?
> * $Library. This is vague, it might be better to add more description?

Reasonable concern. Addressed.

PR #881, opened 2020-09-10
Allow to cancel resolveMainMethod codelens command

testforstephen:
This works with PR https://github.com/microsoft/java-debug/pull/350


testforstephen:
Try to edit a unit test file for a while, below is the profiling result.

resolveMainMethod command is triggered 21 times, but the resolveMainMethodCore only runs 13 times, that means the cancel is hit 8 times.
image

Eskibear (approved):

PR #889, opened 2020-10-09
Fixes #888

### LINKED ISSUE #888: Java multiprject workspaces build fail prompt
When debugging in a workspace with multiple Java projects some may not be building at the moment and so the build will fails.
This triggers a prompt each time you try to debug your project even if it is building correctly.
Until now a configuration flag as been added to disable the build process to this case but if you still want to have your project built you cannot use it.

A better solution would be to add a configuration flag to allow building and proceeding with the debug session even if the build fails.

##### Steps To Reproduce
1. Create workspace
2. Add two Java projects
3. Make one project buildable
4. Make another project un-buildable (introduce a syntax error)
5. Try to debug the buildable project

##### Current Result
Prompts for build failure with following options: "Proceed","Fix...","Cancel"
If "Proceed" selected - allows debugging.

##### Expected Result
If a configuration flag exists with value "Proceed" allow proceeding with debug session.

srfrnk:
Add a configuration value `onBuildFailure` to override the prompt with any of the existing values: `Proceed`, `Fix...`, `Cancel`


package.json
```diff
@@ -636,6 +636,20 @@
           "description": "%java.debugger.configuration.forceBuildBeforeLaunch%",
           "default": true
         },
+        "java.debug.settings.onBuildFailure": {
+          "type": "string",
+          "enum": [
+            "Proceed",
+            "Fix...",
+            "Cancel"
```
testforstephen, line 644: When build fails, pop up `Fix` or `Cancel` silently is not a good behavior. I would suggest the values to be [`askUser`, `proceed`]. Defaults to `askUser` to show current behavior.

Configuration.md
```diff
@@ -1,149 +1,167 @@
...
-* [Modify the settings.json (User Setting)](#modify-the-settingsjson-user-setting)
-  * java.debug.settings.console
-  * java.debug.settings.forceBuildBeforeLaunch
-  * java.debug.settings.hotCodeReplace
-  * java.debug.settings.enableRunDebugCodeLens
-* [FAQ](#faq)
-
-
-The debugger provides two kinds of configuration: *launch.json* and *settings.json* (User Settings), see the [README](https://github.com/Microsoft/vscode-java-debug#options) for the supported configuration list. launch.json is used to control the configuration per debug session, and the user setting is shared by the whole workspace or VS Code.
+
```
testforstephen, line 4: Did you change anything in this file?

package.json
srfrnk: Thx for the suggestion. I actually used "no value" as the indicator for "asking the user". So my implementation right now defaults to the current behaviour if you don't use this settings. Only if you do set this in your config will it override current behavior with the option from the settings... is that not better than having an explicit "do not override" setting?

Configuration.md
srfrnk: I did - added the documentation for the new config setting. I guess auto formatting messed the compare.

srfrnk: e.g. 

line 27:
```md
    -   java.debug.settings.onBuildFailure
```

and...

line 374:
```md
-   `java.debug.settings.onBuildFailure` - Sometimes you may be bothered with the message _"Build failed, do you want to continue?"_, you could use this setting to suppress the message with a specific selection.
```

package.json
testforstephen: You're using null to implicitly indicate something, that's not intuitive. Why not to add it to the enum directly? If you follow the VS Code settings, adding all possible values to the enum is best practice.

Also, please remove `Fix...` and `Cancel` from the settings. When i click debugging, pop up a `Fix` dropdown suddenly or cancel my operation silently is weird.

Configuration.md
testforstephen: The PR should keep the minimum changes. It's better to keep the original formatter, and just add your changes.

srfrnk: done

package.json
srfrnk: I don't agree. I think it's shortsighted to limit the already available options. If someone for some reason explicitly wants to pop the "Fix..." UI automatically I don't think it should be limited just because it might look "weird" at this point. But I'll just go ahead and change this to move this forward.

Configuration.md
```diff
@@ -329,27 +330,29 @@ In some cases, you may want to start your program with the external builder and
...
+  - `manual` - Click the toolbar to apply the changes.
   ![hcr](https://user-images.githubusercontent.com/14052197/67256313-f5697600-f4b8-11e9-9db6-54540b6350ad.png)
   - `auto` - Automatically apply the changes after saved.
   - `never` - Never apply the changes.
 
 - `java.debug.settings.enableRunDebugCodeLens` - Enable the code lens provider for the *Run* and *Debug* buttons over main entry points, defaults to `true`.
-  - `true` - Show the code lens.  
+  - `true` - Show the code lens.
   ![codelens](https://user-images.githubusercontent.com/14052197/67256585-83922c00-f4ba-11e9-883f-2b3de3db2dfa.png)
-  - `false` - Show the Run/Debug link in the hover.
```
testforstephen, line 341: Again, please keep the tailing 2 spaces, which is used for newline.

srfrnk: All formatting adjusted to minimal changes.

testforstephen (approved):

testforstephen:
@srfrnk thanks for the contribution.

srfrnk:
thanks @testforstephen. will be waiting for the version distro!

PR #891, opened 2020-10-10
Provide a user setting to set default vmArgs

### LINKED ISSUE #220: Provide a way to set default vmArgs
It's nice that I can set "vmArgs" in a launch configuration, but it's tedious to have to type it again and again for each project I start, and each launcher I create. Some way to set the defaults (and view it in case I want to change it) would be good.

### LINKED ISSUE #876: Add setting to enable asserts.
Currently when I run the debugger, my asserts are not being triggered.

It would be nice if there was a way I could turn on asserts, potentially in an extension setting or something.

testforstephen:
Close #220
Close #876

### DIFF
--- a/package.json
+++ b/package.json
@@ -730,6 +730,11 @@
           "description": "%java.debugger.configuration.jdwp.requestTimeout.description%",
           "default": 3000,
           "minimum": 100
+        },
+        "java.debug.settings.vmArgs": {
+          "type": "string",
+          "description": "%java.debugger.configuration.vmArgs.description%",
+          "default": ""
         }
       }
     }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -730,6 +730,11 @@
           "description": "%java.debugger.configuration.jdwp.requestTimeout.description%",
           "default": 3000,
           "minimum": 100
+        },
+        "java.debug.settings.vmArgs": {
```
jdneo, line 734: Do you think if it is necessary to set it as `machine` scope?

testforstephen: Seems not, because `vmArgs` is not auto executed by the debugger. It's like `vmArgs` in launch.json, need user to click debug to consume it.

jdneo: I see, that's reasonable.

jdneo (approved):

PR #898, opened 2020-10-28
Make the run buttons in editor toolbar more context-aware

jdneo:
- If the file has a main() method, run it
- If the file is a test file, delegate to Test Runner
- If neither has a main() method nor is a test file, try to launch as a project
- If both has a main() method and is a test file, let the user to choose

Signed-off-by: Sheng Chen <sheche@microsoft.com>

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -268,12 +268,7 @@ async function runJavaFile(uri: vscode.Uri, noDebug: boolean) {
     const canRunTests: boolean = await canDelegateToJavaTestRunner(uri);
 
     if (!hasMainMethods && !canRunTests) {
-        const workspaceFolder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(uri);
-        if (!workspaceFolder) {
-            vscode.window.showErrorMessage(`Failed to run the project because the file: ${uri.fsPath} does not belongs to the current workspace.`);
-            return;
-        }
-        const mainClasses: IMainClassOption[] = await utility.searchMainMethods(workspaceFolder.uri);
+        const mainClasses: IMainClassOption[] = await utility.searchMainMethods();
         await launchMain(mainClasses, uri, noDebug);
     } else if (hasMainMethods && !canRunTests) {
         await launchMain(mainMethods, uri, noDebug);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -264,6 +264,50 @@ async function runJavaFile(uri: vscode.Uri, noDebug: boolean) {
         throw ex;
     }
 
+    const hasMainMethods: boolean = mainMethods.length > 0;
+    const canRunTests: boolean = await canDelegateToJavaTestRunner(uri);
+
+    if (!hasMainMethods && !canRunTests) {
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -264,6 +264,50 @@ async function runJavaFile(uri: vscode.Uri, noDebug: boolean) {
...
+    const hasMainMethods: boolean = mainMethods.length > 0;
+    const canRunTests: boolean = await canDelegateToJavaTestRunner(uri);
+
+    if (!hasMainMethods && !canRunTests) {
+        const workspaceFolder: vscode.WorkspaceFolder | undefined = vscode.workspace.getWorkspaceFolder(uri);
+        if (!workspaceFolder) {
+            vscode.window.showErrorMessage(`Failed to run the project because the file: ${uri.fsPath} does not belongs to the current workspace.`);
+            return;
+        }
+        const mainClasses: IMainClassOption[] = await utility.searchMainMethods(workspaceFolder.uri);
```
testforstephen, line 276: Can we just search the main method from the whole workspace directly instead of limiting to the belonging root?

jdneo: Update the utility method to make it possible to search all main classes in the workspace. Another question, do you think we need to differentiate the "mains" by appending the belonging project name in the list?


image

testforstephen: yes. if it falls back to run the main from the workspace, adding project name in the list would help.

jdneo: New look:

WeChat Screenshot_20201030135941

src/utility.ts
```diff
@@ -248,3 +250,24 @@ export async function waitForStandardMode(): Promise<boolean> {
...
+        return await vscode.window.withProgress<IMainClassOption[]>(
+            { location: vscode.ProgressLocation.Window },
+            async (p) => {
+                p.report({ message: "Searching main classes..." });
+                if (_.isEmpty(uris)) {
+                    uris = vscode.workspace.workspaceFolders.map((folder) => folder.uri);
+                }
+                const mainClasses: IMainClassOption[] = [];
+                for (const uri of uris) {
+                    mainClasses.push(...await resolveMainClass(uri));
```
testforstephen, line 265: Calling search engine multiple times is not efficient. In current PR, i didn't see there is requirement to search for multiple uris at one time. I suggest to keep passing at most 1 uri in this method. If there is requirement in future to search for multiple uris, we can improve the Java side logic for that.

jdneo: But what if the workspace has multiple folders? calling `searchMainMethods()` multiple times?

testforstephen: If you didn't pass uri in this method, it means searching in all folders.

If you look at the Java implementation, what it did is searching all results first. and if there is a uri given, then filter the result with the uri, otherwise, return all.

testforstephen: https://github.com/microsoft/java-debug/blob/master/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java#L77

jdneo: I see

testforstephen (approved):

testforstephen:
thanks @jdneo

PR #918, opened 2020-12-08
Enable strict null checks in tsconfig

xqzlgy2:
This PR enabled TS compiler check for `strictNullChecks`, also fixed existing errors.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -317,7 +317,7 @@ async function launchMain(mainMethods: IMainClassOption[], uri: vscode.Uri, noDe
         return;
     }
 
-    await startDebugging(pick.mainClass, pick.projectName!, uri, noDebug);
+    await startDebugging(pick.mainClass, pick.projectName || "", uri, noDebug);
 }
 
 async function runJavaProject(node: any, noDebug: boolean) {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -165,7 +165,7 @@ function specifyProgramArguments(context: vscode.ExtensionContext): Thenable<str
 }
 
 async function applyHCR(hcrStatusBar: NotificationBar) {
-    const debugSession: vscode.DebugSession = vscode.debug.activeDebugSession;
+    const debugSession: vscode.DebugSession | undefined = vscode.debug.activeDebugSession;
     if (!debugSession) {
         return;
     }
@@ -246,7 +246,7 @@ async function runJavaFile(uri: vscode.Uri, noDebug: boolean) {
         throw ex;
     }
 
-    const activeEditor: vscode.TextEditor = vscode.window.activeTextEditor;
+    const activeEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
     if (!uri && activeEditor && _.endsWith(path.basename(activeEditor.document.fileName), ".java")) {
         uri = activeEditor.document.uri;
     }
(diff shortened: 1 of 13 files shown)


src/extension.ts
```diff
@@ -317,7 +317,7 @@ async function launchMain(mainMethods: IMainClassOption[], uri: vscode.Uri, noDe
         return;
     }
 
-    await startDebugging(pick.mainClass, pick.projectName, uri, noDebug);
+    await startDebugging(pick.mainClass, pick.projectName!, uri, noDebug);
```
testforstephen, line 320: pick.projectName!

What is the purpose of the ending **!**

xqzlgy2: Because the second parameter of `startDebugging` could not be undefined, but `pick.projectName` could be, since it's  optional. How about change projectName in `IMainClassOption` to required property?

testforstephen: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

x! will throw exception if x is null or undefined. This is not we want. I would change it to 
`await startDebugging(pick.mainClass, pick.projectName || "", uri, noDebug);`

xqzlgy2: OK. BTW, I didn't do this because I don't know what is a proper default value.

src/utility.ts
```diff
@@ -55,11 +55,11 @@ function logMessage(message: ILoggingMessage): void {
         }
         sendError(error);
     } else {
-        sendInfo(null, { message: message.message });
+        sendInfo("", { message: message.message });
     }
 
     // Deprecated
-    logger.log(message.type, { message: message.message, stack: message.stack });
+    logger.log(message.type, { message: message.message, stack: message.stack ? message.stack : "" });
```
testforstephen, line 62: message.stack ? message.stack : "" => message.stack || ""

src/utility.ts
```diff
@@ -194,11 +194,14 @@ export function isJavaExtEnabled(): boolean {
 
 export function isJavaExtActivated(): boolean {
     const javaExt = vscode.extensions.getExtension(JAVA_EXTENSION_ID);
-    return javaExt && javaExt.isActive;
+    if (javaExt && javaExt.isActive) {
+        return true;
+    }
+    return false;
```
testforstephen, line 200: what's the problem with the old usage?

src/mainClassPicker.ts
```diff
@@ -88,16 +88,21 @@ class MainClassPicker {
...
-        const mostRecentlyUsedOption: IMainClassOption = (options.length && this.contains(options[0])) ? options[0] : undefined;
+        const mostRecentlyUsedOption: IMainClassOption | undefined = (options.length && this.contains(options[0])) ? options[0] : undefined;
         const isMostRecentlyUsed = (option: IMainClassOption): boolean => {
-            return mostRecentlyUsedOption
-                && mostRecentlyUsedOption.mainClass === option.mainClass
+            if (mostRecentlyUsedOption) {
+                return mostRecentlyUsedOption.mainClass === option.mainClass
                 && mostRecentlyUsedOption.projectName === option.projectName;
+            }
+            return false;
```
testforstephen, line 97: using `!!mostRecentlyUsedOption && mostRecentlyUsedOption.mainClass === option.mainClass && mostRecentlyUsedOption.projectName === option.projectName` ?

testforstephen (changes requested):

src/utility.ts
xqzlgy2: `Type 'boolean | undefined' is not assignable to type 'boolean'`
How about `return !!javaExt && javaExt.isActive;` ?

testforstephen (approved):

PR #919, opened 2020-12-08
Use a progress reporter to hint the current build status when run or debug a program

testforstephen:
This PR includes:
- Use a consistent progress reporter to report the build status before running/debugging an application. 
By default, progress notifications are used to display status. You can also use the status bar to display status by enabling user settings `java.silentNotification`.

- Expose an extension API for test runner to use the same progress reporter. In addition, when third-party extensions such as test runner call the startDebugging API to start the debugger, they can add an attribute `__progressId` to the launchConfiguration so that the debugger can use the same progress reporter to report progress.
```ts
// The debugger will export the progressProvider instance to others for reuse.
{
  progressProvider: IProgressProvider
}

export interface IProgressReporter {
    /**
     * Returns the id of the progress reporter.
     */
    getId(): string;

    /**
     * Reports a progress message update.
     * @param message the message to update
     * @param increment use `increment` to report discrete progress. Each call with a `increment`
     *                  value will be summed up and reflected as overall progress until 100% is reached.
     *                  Note that currently only `ProgressLocation.Notification` is capable of showing
     *                  discrete progress.
     */
    report(message: string, increment?: number): void;

    /**
     * Shows the progress reporter.
     */
    show(): void;

    /**
     * Hides the progress reporter.
     * @param onlyNotifications only hide the progress reporter when it's shown as notification.
     */
    hide(onlyNotifications?: boolean): void;

    /**
     * Returns whether the progress reporter has been cancelled or completed.
     */
    isCancelled(): boolean;

    /**
     * Notifies the work is done that is either the task is completed or the user has cancelled it.
     */
    done(): void;

    /**
     * The CancellationToken to monitor if the progress reporter has been cancelled by the user.
     */
    getCancellationToken(): CancellationToken;

    /**
     * Disposes the progress reporter if the observed token has been cancelled.
     * @param token the cancellation token to observe
     */
    observe(token?: CancellationToken): void;
}

export interface IProgressProvider {
    /**
     * Creates a progress reporter.
     * @param jobName the job name
     * @param progressLocation The location at which progress should show, defaults to `ProgressLocation.Notification`.
     * @param cancellable Controls if a cancel button should show to allow the user to cancel the progress reporter,
     *                    defaults to true. Note that currently only `ProgressLocation.Notification` is supporting
     *                    to show a cancel button.
     */
    createProgressReporter(jobName: string, progressLocation?: ProgressLocation | { viewId: string }, cancellable?: boolean): IProgressReporter;

    /**
     * Returns the progress reporter with the progress id.
     * @param progressId the progress id
     */
    getProgressReporter(progressId: string): IProgressReporter | undefined;
}

```

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -175,7 +175,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         if (!progressReporter && config.__progressId) {
             return undefined;
         } else if (!progressReporter) {
-            progressReporter = progressProvider.createProgressReporter(config.noDebug ? "Run" : "Debug", vscode.ProgressLocation.Notification, true);
+            progressReporter = progressProvider.createProgressReporter(config.noDebug ? "Run" : "Debug");
         }
 
         progressReporter.observe(token);
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -17,7 +17,7 @@ import { logger, Type } from "./logger";
 import { mainClassPicker } from "./mainClassPicker";
 import { resolveJavaProcess } from "./processPicker";
 import { IProgressReporter } from "./progressAPI";
-import { progressReporterManager } from "./progressImpl";
+import { progressProvider } from "./progressImpl";
 import * as utility from "./utility";
 
 const platformNameMappings: {[key: string]: string} = {
@@ -90,7 +90,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
 
     private provideDebugConfigurationsAsync(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken) {
         return new Promise(async (resolve, _reject) => {
-            const progressReporter = progressReporterManager.create("Create launch.json", true);
+            const progressReporter = progressProvider.createProgressReporter("Create launch.json", vscode.ProgressLocation.Window);
             progressReporter.observe(token);
             const defaultLaunchConfig = {
                 type: "java",
@@ -171,11 +171,11 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
 
     private async resolveAndValidateDebugConfiguration(folder: vscode.WorkspaceFolder | undefined, config: vscode.DebugConfiguration,
                                                        token?: vscode.CancellationToken) {
-        let progressReporter = progressReporterManager.get(config.__progressId);
+        let progressReporter = progressProvider.getProgressReporter(config.__progressId);
         if (!progressReporter && config.__progressId) {
             return undefined;
         }
-        progressReporter = progressReporter || progressReporterManager.create(config.noDebug ? "Run" : "Debug");
+        progressReporter = progressReporter || progressProvider.createProgressReporterForPreLaunchTask(config.noDebug ? "Run" : "Debug");
         progressReporter.observe(token);
         if (progressReporter.isCancelled()) {
             return undefined;
@@ -221,6 +221,8 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     if (!proceed) {
                         return undefined;
                     }
+
+                    progressReporter.show();
                 }
 
                 if (progressReporter.isCancelled()) {
@@ -353,7 +355,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
             utility.showErrorMessageWithTroubleshooting(utility.convertErrorToMessage(ex));
             return undefined;
         } finally {
-            progressReporter.cancel();
+            progressReporter.done();
         }
     }
 
@@ -388,7 +390,7 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
             if (currentFile) {
                 const mainEntries = await lsPlugin.resolveMainMethod(vscode.Uri.file(currentFile));
                 if (mainEntries.length) {
-                    progressReporter.report("Select mainClass", "Selecting the main class to run...");
+                    progressReporter.hide(true);
                     return mainClassPicker.showQuickPick(mainEntries, "Please select a main class you want to run.");
                 }
             }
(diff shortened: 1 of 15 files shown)


testforstephen:
A private vsix to test the progress reporter feature. https://github.com/testforstephen/vscode-test/blob/master/vscode-java-debug-0.29.1.vsix

progressReporter

jdneo:
I have one question: When do we need to manually call `progressReporterManager.remove()`?

testforstephen:
> I have one question: When do we need to manually call progressReporterManager.remove()?

Good question, you don't need call it. It's supposed to be used by progressReporter.cancel().

src/progressImpl.ts
```diff
@@ -0,0 +1,136 @@
...
+        this._disposables.push(this._cancelEventEmitter);
+        this._disposables.push(this._progressEventEmitter);
+        this._disposables.push(this._tokenSource);
+    }
+
+    public getId(): string {
+        return this._id;
+    }
+
+    public report(subTaskName: string, detailedMessage: string): void {
```
jdneo, line 55: Change to `detailedMessage?: string` since it's an optional param?

src/extension.ts
```diff
@@ -329,35 +341,53 @@ async function runJavaProject(node: any, noDebug: boolean) {
...
-    const debugConfig = existConfig || {
-        type: "java",
-        name: `Launch - ${mainClass.substr(mainClass.lastIndexOf(".") + 1)}`,
-        request: "launch",
-        mainClass,
-        projectName,
-    };
-    debugConfig.noDebug = noDebug;
-    vscode.debug.startDebugging(workspaceFolder, debugConfig);
+        progressReporter.report("Select mainClass", "Selecting the main class to run...");
```
jdneo, line 358: Here is my first thinking when I saw the message here.

Basically the messages are used to telling the user what the `extension` is doing right now. So it seems that it's telling the user the extension is selecting the main class, but actually it's waiting for the user to select.

src/configurationProvider.ts
```diff
@@ -86,44 +88,56 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                if (ex instanceof utility.JavaExtensionNotEnabledError) {
+                    utility.guideToInstallJavaExtension();
+                } else {
+                    // tslint:disable-next-line
+                    console.error(ex);
+                }
+
+                resolve([defaultLaunchConfig]);
+            } finally {
+                progressReporter.cancel();
```
jdneo, line 139: Does that mean that the invocation of `progressReporter.cancel()` is required as its lifecycle method? If that is the case, maybe calling it as `.done()` is better.

src/extension.ts
```diff
@@ -300,24 +302,34 @@ async function canDelegateToJavaTestRunner(uri: vscode.Uri): Promise<boolean> {
     return (await vscode.commands.getCommands()).includes("java.test.editor.run");
 }
 
-async function launchTesting(uri: vscode.Uri, noDebug: boolean): Promise<void> {
-    noDebug ? vscode.commands.executeCommand("java.test.editor.run", uri) : vscode.commands.executeCommand("java.test.editor.debug", uri);
+function launchTesting(uri: vscode.Uri, noDebug: boolean, progressReporter: IProgressReporter) {
+    const command: string = noDebug ? "java.test.editor.run" : "java.test.editor.debug";
+    vscode.commands.executeCommand(command, uri, progressReporter);
+    if (compareVersions(getTestExtensionVersion(), "0.26.1") <= 0) {
```
jdneo, line 308: the run test from editor feature start from `0.25.0`.

So I think it should be

```js
if (compareVersions(getTestExtensionVersion(), "0.25.0") < 0) {
    throw new utility.OperationCancelledError("");
}
```

testforstephen: Below is the execution flow when clicking run button:
the debugger pre-check the file info (show progress reporter)  -> trigger run command in test runner -> trigger configuration provider (show progress reporter)

If the test runner doesn't support progress reporter parameter at the second stage, the debugger needs to cancel the progress reporter at the stage 1, otherwise there is no way to terminate it.

so this is just used to check whether the test runner supports progress reporter.

jdneo: OK

src/configurationProvider.ts
testforstephen: `.done()` makes sense.

testforstephen:
Here is a new private vsix https://github.com/testforstephen/vscode-test/blob/master/vscode-java-debug-0.29.2.vsix

src/configurationProvider.ts
```diff
@@ -227,17 +229,17 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     return undefined;
                 }
                 if (!config.mainClass) {
-                    progressReporter.report("Resolve mainClass", "Resolving main class...");
+                    progressReporter.report("Resolving mainClass...");
```
jdneo, line 232: `Resolving main class...`

And don't forget to update all the other appearance for this message

src/progressImpl.ts
```diff
@@ -0,0 +1,158 @@
...
+    private _progressLocation: ProgressLocation | { viewId: string };
+    private _cancellable: boolean = false;
+
+    private _message: string;
+    private _increment: number | undefined;
+    private _isShown: boolean;
+
+    private _tokenSource = new CancellationTokenSource();
+    private _statusBarItem: StatusBarItem | undefined;
+    private _cancelProgressEventEmitter: EventEmitter<any>;
```
jdneo, line 22: any -> void

src/progressImpl.ts
```diff
@@ -0,0 +1,158 @@
...
+    private _progressEventEmitter: EventEmitter<any>;
```
jdneo, line 23: any -> void

src/progressAPI.ts
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Creates a progress reporter.
+     * @param jobName the job name
+     * @param progressLocation The location at which progress should show
+     * @param cancellable Controls if a cancel button should show to allow the user
+     *                    to cancel the progress reporter. Note that currently only
+     *                    `ProgressLocation.Notification` is supporting to show a cancel
+     *                    button.
+     */
+    createProgressReporter(jobName: string, progressLocation: ProgressLocation | { viewId: string }, cancellable?: boolean): IProgressReporter;
```
jdneo, line 65: We can provide default value for the `progressLocation` and `cancellable` here

jdneo (approved):

PR #943, opened 2021-02-01
Prepare new release 0.31.0

testforstephen:


package.json
```diff
@@ -47,7 +47,7 @@
       }
     ],
     "javaExtensions": [
-      "./server/com.microsoft.java.debug.plugin-0.29.0.jar"
+      "./server/com.microsoft.java.debug.plugin-0.30.0.jar"
```
jdneo, line 50: Is `30` here on purpose?

testforstephen: yes. In latest debugger release, we didn't modify the code in java-debug. So there is one version gap between them.

jdneo (approved):

Eskibear (approved):

PR #977, opened 2021-04-12
Use InlineValuesProvider to provide inline debugging feature

testforstephen:
This requires https://github.com/microsoft/java-debug/pull/368

### DIFF
--- a/src/JavaInlineValueProvider.ts
+++ b/src/JavaInlineValueProvider.ts
@@ -1,0 +1,76 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as compareVersions from "compare-versions";
+import { debug, InlineValue, InlineValueContext, InlineValueEvaluatableExpression, InlineValuesProvider, InlineValueText, InlineValueVariableLookup,
+    Position, Range, TextDocument, version } from "vscode";
+import { InlineKind, InlineVariable, LSPRange, resolveInlineVariables } from "./languageServerPlugin";
+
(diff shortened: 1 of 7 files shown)


src/JavaInlineValueProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+            character: range.start.character,
+        },
+        end: {
+            line: range.end.line,
+            character: range.end.character,
+        },
+    };
+}
+
+function asClientRange(range: LSPRange) {
```
jdneo, line 74: There is a same utility in `vscode-languageclient/lib/protocolConverter`, if you want to use.

_// Since we are using webpack so introduce a new module is fine_

src/JavaInlineValueProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+                declaringClass: variable.declaringClass,
+            };
+        });
+        let resolvedVariables: any;
+        if (unresolvedVariables.length && debug.activeDebugSession) {
+            const response = await debug.activeDebugSession.customRequest("inlineValues", {
+                frameId: context.frameId,
+                variables: unresolvedVariables,
+            });
+            resolvedVariables = response && response.variables ? response.variables : undefined;
```
jdneo, line 36: `resolvedVariables = response?.variables`

src/JavaInlineValueProvider.ts
testforstephen: Nice.

jdneo (approved):

testforstephen:
In VS Code 1.55.0, you have to manually enable the inline values feature via user setting `"debug.inlineValues"`.
In 1.56.0 and above, it's enabled by default.

testforstephen:
image

image

PR #987, opened 2021-04-26
Clean up deprecated logs

testforstephen:


testforstephen:
The deprecated logs are all covered by telemetry wrapper at [0.28.0 release](https://github.com/microsoft/vscode-java-debug/blob/master/CHANGELOG.md#0280---2020-08-27) (2020-08-27). It has been half years ago, it's time to cleanup it.

src/logger.ts
```diff
@@ -1,63 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license.
-
-import * as fs from "fs";
-import * as vscode from "vscode";
-import TelemetryReporter from "vscode-extension-telemetry";
```
jdneo, line 6: The dependency of `vscode-extension-telemetry` can also be removed in package.json

Eskibear (approved):

src/logger.ts
testforstephen: done.

PR #988, opened 2021-04-27
adopt workspaceTrust in package.json

Eskibear:


Eskibear:
part of https://github.com/microsoft/vscode-java-pack/issues/593

package.json
```diff
@@ -7,6 +7,11 @@
   "preview": true,
   "aiKey": "67d4461e-ccba-418e-8082-1bd0acfe8516",
   "icon": "logo.png",
+  "enableProposedApi": true,
```
testforstephen, line 10: If I remember correctly, VS Code would prevent an extension with enableProposedApi tag from being released to marketplace.

testforstephen: https://code.visualstudio.com/updates/v1_31#_proposed-extension-apis

But i saw python extension added this tag. Not sure how they did that.

Eskibear: > But i saw python extension added this tag

Yes, I checked python extension downloaded from marketplace, and it's in package.json. I'm going to have a try and I'll let you know whether I can sucsessfully upload the extension.

Eskibear: > Note that you cannot publish an extension that uses a proposed API. We may likely make breaking changes in the next release and we never want to break existing extensions

I guess it doesn't matter to add `enableProposedApi: true`, as long as we don't actually **use** a proposed API.

Eskibear: Tried. It doesn't block any feature. But in console there will be an error log, which might be confusing for users. I'll remove this line for the moment.

testforstephen (approved):

PR #1046, opened 2021-09-07
Remove test projects from main branch

testforstephen:
Remove these test projects from `main` branch for security issue. You still can view it at a side branch `test_samples`.


jdneo (approved):
Don't forget to add a protect rule for the branch `test_samples` to prevent removing accidentally

PR #1082, opened 2021-11-22
Update Troubleshooting_encoding.md

jdneo:
I think we can use some html tags to show the picture caption in a more fancy way. The title in the screenshot then can be removed.

Meanwhile, another thought about the change locale screenshot. This is Windows 7 style dialog. I'm not quite sure if it's suitable here.


testforstephen:
Here is a new screenshot without title words.
image


> Meanwhile, another thought about the change locale screenshot. This is Windows 7 style dialog. I'm not quite sure if it's suitable here.

hmm, i captured it from my Windows 10 machine.

jdneo:
> hmm, i captured it from my Windows 10 machine.

Ok, then that's fine.

testforstephen (approved):

PR #1123, opened 2022-02-14
Add check to determine if javaExec is specified in the

Rod-Christensen:
This fork allows you to specify the javaExec to use for executing the java virtual machine. It can be specified in your launch configuration options, or, if not specified, will use the current behavior of resolving.

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -284,8 +284,10 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     });
                 }
 
-                config.javaExec = await lsPlugin.resolveJavaExecutable(config.mainClass, config.projectName);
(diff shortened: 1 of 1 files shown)


Rod-Christensen:
I really don't know what to do from here... I guess this is just waiting for approval?

testforstephen:
https://github.com/microsoft/vscode-java-debug/blob/a3f48af1513c9b293a50b465dc43d9315b58bc58/package.json#L377-L385

Usually when adding a new property to launch configuration, we update package.json to include the new property schema definition as well. See the sample for `mainClass` above.  This can provide IntelliSense when you type the property in launch.json.

Also remember to update it to README. https://github.com/microsoft/vscode-java-debug/blob/main/README.md#launch

PR #1134, opened 2022-03-14
Update VS Code requirement to `1.65`

CsCherrYY:
since we will support lazy loading in https://github.com/microsoft/java-debug/pull/401, the requirement of this feature is introduced in VS Code 1.65, see: https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_65.md#lazy-variables

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "@types/mocha": "^9.1.0",
         "@types/node": "^14.18.12",
         "@types/uuid": "^8.3.4",
-        "@types/vscode": "1.57.0",
+        "@types/vscode": "^1.65.0",
         "mocha": "^9.2.2",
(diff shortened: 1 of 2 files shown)


CsCherrYY:
> Since the lazy variable implementation is compatible with old vscode client, it's not necessary to force updating VS Code version. WDYT?

Agree that it's not necessary, but if you run this version in an old client, you will see a similar behavior like https://github.com/microsoft/vscode/issues/134450#issuecomment-971867869, for the getter of address. it will introduce an extra node so I'm not sure if it's a good UX. It's up to you.

testforstephen:
@CsCherrYY Could you pls resolve the conflicts?

testforstephen (approved):

Issue #710, opened 2019-11-12
Output logpoints to debug console


jvalkeal:
Took a while to figure out that logpoints output simply goes to process(java process to debugged) output and not vscode's debug console as it's shown in docs. It'd be extremely helpful especially with remote debugging to get those logpoint messages back via debug protocol if that's possible.

If I'm right this `System.out` happens in this companion project [JdtEvaluationProvider.java#L225](https://github.com/microsoft/java-debug/blob/6633a55079db64d2893092d49b04a606f183b816/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java#L225)

sbj42:
Another place to look is [SetBreakpointsRequestHandler.java#L219](https://github.com/microsoft/java-debug/blob/main/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java#L219).  There is some code there to notify the user if an error occurred in the logpoint evaluation by sending a `UserNotificationEvent`.  That event is picked up in TypeScript at [extension.ts#L126](https://github.com/microsoft/vscode-java-debug/blob/main/src/extension.ts#L126).  If there was no error, this code doesn't do anything.

I'm just guessing, but maybe a fix would be for `JdtEvaluationProvider.logMessageToExpression()` to generate an expression which returns the formatted log message, then `SetBreakpointsRequestHandler.handleEvaluationResult()` would extract the string and send it in a new kind of event, say a "logpoint" event.  Then `extension.ts` would handle that event and write to the debug console.

PR #1139, opened 2022-03-16
Send logpoint messages to the debug console

sbj42:
This change receives logpoint messages as "logpoint" custom events from the java-debug java extension, and then prints them to the current active debug console.

Together with microsoft/java-debug#402, this would address microsoft/vscode-java-debug#710


sbj42:
Here's a demonstration of this change:
java-logpoint

testforstephen:
If you use the standard OutputEvent to print logpoint message, this PR is not needed.

PR #1168, opened 2022-05-24
Enable Java terminal link provider on all terminals

testforstephen:
This is a follow-up of PR https://github.com/microsoft/java-debug/pull/414 and https://github.com/redhat-developer/vscode-java/issues/2463.

Enabling Java terminal link providers on all terminals can also benefit the scenarios of running Maven command in VS Code terminal.

### DIFF
--- a/src/terminalLinkProvider.ts
+++ b/src/terminalLinkProvider.ts
@@ -15,10 +15,6 @@ export class JavaTerminalLinkProvder implements TerminalLinkProvider<IJavaTermin
      * @return A list of terminal links for the given line.
      */
     public provideTerminalLinks(context: TerminalLinkContext, _token: CancellationToken): ProviderResult<IJavaTerminalLink[]> {
-        if (context.terminal.name !== "Java Debug Console" && context.terminal.name !== "Java Process Console") {
-            return [];
-        }
-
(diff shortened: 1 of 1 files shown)


CsCherrYY:
Is there any potential issue if we remove this restriction? e.g. conflicts in terminals created by another extension.

If so, we can also keep this restriction for the known usages via a whitelist for example.

jdneo (approved):

Eskibear (approved):
LGTM

testforstephen:
> 

Yes, this is a good concern. VS Code supports returning a list of terminal links for each line. From our experience with existing providers, VS Code should merge results from different extensions.

See its API definition.
image

Eskibear:
> Is there any potential issue if we remove this restriction? e.g. conflicts in terminals created by another extension.
> 
> If so, we can also keep this restriction for the known usages via a whitelist for example.

I think the restriction was designed so. But I prefer to remove the restriction as it doesn't seem to do any harm. We can revisit it if there's any case that links are overlap with popular extensions.

testforstephen:
Agree, we can roll out first. If there are any conflicts with other popular extensions, then we can limit its scope later.

CsCherrYY:
Agree. Let's remove this for now.

PR #1198, opened 2022-07-30
Add support to specify java executable

gayanper:
This new optional option will allow to specify java executable path to
use to execute the program.

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -295,7 +295,17 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     });
                 }
 
-                config.javaExec = await lsPlugin.resolveJavaExecutable(config.mainClass, config.projectName);
+                if (_.isEmpty(config.javaExec)) {
+                    config.javaExec = await lsPlugin.resolveJavaExecutable(config.mainClass, config.projectName);
+                } else {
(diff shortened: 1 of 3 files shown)


gayanper:
@testforstephen i would like to get your feddback on this change.

src/configurationProvider.ts
```diff
@@ -295,7 +295,25 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
                     });
                 }
 
-                config.javaExec = await lsPlugin.resolveJavaExecutable(config.mainClass, config.projectName);
+                if (_.isEmpty(config.javaRuntime)) {
```
testforstephen, line 298: I would add the new config name as `javaExec`, if it's not specified, then the debugger will auto resolve it from project runtime. In addition, why not pass the Java path directly? Specifying a URI is not user friendly.

gayanper: Name suggestion is good. Will change that. The reason for using a URI was to add support for configured jvm runtimes by name later. In vscode java you can configure runtimes, so i was thinking to add support later for that as well through a custom uri. Otherwise there was no special reason.

testforstephen: In my experience, asking the user to manually specify a URI is very error-prone. I would prefer it to be straightforward.

testforstephen (approved):
LGTM

PR #1204, opened 2022-08-10
Send telemetry when adding source paths

jdneo:
Signed-off-by: sheche <sheche@microsoft.com>

### DIFF
--- a/src/debugCodeLensProvider.ts
+++ b/src/debugCodeLensProvider.ts
@@ -214,6 +214,7 @@ async function addToClasspath(uri: vscode.Uri): Promise<boolean> {
     if (parentPath === parentUri.fsPath) {
         parentPath = path.basename(parentFsPath);
     }
+    sendInfo("", {operationName: "notOnClasspath"});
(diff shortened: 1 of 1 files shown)


src/debugCodeLensProvider.ts
```diff
@@ -219,6 +219,7 @@ async function addToClasspath(uri: vscode.Uri): Promise<boolean> {
     if (ans === "Skip") {
         return true;
     } else if (ans === "Add to Source Path") {
+        sendInfo("", {operationName: "addToSourcePath"});
```
testforstephen, line 222: It might be better to record each occurrence of `notOnClasspath` before Line 217 as well.

`sendInfo("", { operationName: "notOnClasspath" })`

jdneo: Will

```
sendInfo(operationId, {
            buildErrorTypes: “otnOnClasspath”,
            buildErrorCount: 1,
        });
```

better?

testforstephen (approved):

PR #1208, opened 2022-08-30
Add a setting to control whether to enable async jdwp

testforstephen:
This PR relates to https://github.com/microsoft/java-debug/pull/436

### DIFF
--- a/src/configurationProvider.ts
+++ b/src/configurationProvider.ts
@@ -737,6 +737,11 @@ async function updateDebugSettings(event?: vscode.ConfigurationChangeEvent) {
             const exceptionFilters = {
                 skipClasses: await substituteFilterVariables(debugSettingsRoot.settings.exceptionBreakpoint.skipClasses),
             };
+
+            let asyncJDWP: string = debugSettingsRoot.settings.jdwp.async;
+            if (asyncJDWP === "auto" && vscode?.env?.appName === "Visual Studio Code - Insiders") {
+                asyncJDWP = "on";
+            }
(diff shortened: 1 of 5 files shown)


src/configurationProvider.ts
```diff
@@ -737,6 +737,11 @@ async function updateDebugSettings(event?: vscode.ConfigurationChangeEvent) {
             const exceptionFilters = {
                 skipClasses: await substituteFilterVariables(debugSettingsRoot.settings.exceptionBreakpoint.skipClasses),
             };
+
+            let asyncJDWP: string = debugSettingsRoot.settings.jdwp.async;
+            if (asyncJDWP === "auto" && vscode?.env?.appName === "Visual Studio Code - Insiders") {
```
jdneo, line 742: `vscode.env?.appName` I think the `?` after `vscode` is unnecessary?

testforstephen: nice.

jdneo (approved):

Eskibear:
Will this setting appear on settings page for vscode stable editions? If so, shall we explicitly note "for Insiders only" also in this setting's description instead of "experimental" to avoid confusion? Because for stable vscode, I also see some native "experimental" settings that works.

testforstephen:
> Will this setting appear on settings page for vscode stable editions? If so, shall we explicitly note "for Insiders only" also in this setting's description instead of "experimental" to avoid confusion? Because for stable vscode, I also see some native "experimental" settings that works.

This is not for Insiders only, just enable it on Insiders by default. For stable, you still can manually opt in.

Eskibear:
> This is not for Insiders only, just enable it on Insiders by default. For stable, you still can manually opt in.

Sorry for misunderstanding. Then it looks good to me.

PR #1263, opened 2022-11-17
Only pop-up build errors notification when the errors are on the running project's classpath

### LINKED ISSUE #1214: Optimize the build error report of "Build failed, do you want to continue?"
**Customer verbatim:**
"the java debugger extentions sometimes gives errors but there is not any errors" 
"Sometimes it gets debugger errors" 
"Everything is good but when we run the code it have a problem it shows first run the code.  Please fix it then it will run the code."

testforstephen:
Fixes #1214
Requires https://github.com/microsoft/java-debug/pull/459

- For a build-tool managed project, will check if the running project and its references projects contain build errors. If so, report build failure. Otherwise, return succeed.
- For unmanaged folder, only check if the running main class contains build errors. If so, report build failure. Otherwise, return succeed.
- Add a "Always Continue" button in case users want to always continue on build failure.

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -11,6 +11,7 @@ import * as utility from "./utility";
 
 const JAVA_DEBUG_CONFIGURATION = "java.debug.settings";
 const ON_BUILD_FAILURE_PROCEED = "onBuildFailureProceed";
+const CANCELLED_CODE = -32800;
 
 enum CompileWorkspaceStatus {
     FAILED = 0,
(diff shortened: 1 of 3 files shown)


jdneo (approved):
LGTM.

@CsCherrYY I remember there is a copy for the build stuff in project manager when exporting jar files. Could you please check if we need to update there as well?

CsCherrYY:
And the logics at the project manager side can be reviewed. E.g., we can only build the given project, instead of the whole workspace, than we can benefit from this change.

jdneo:
JLS supports build specific project: https://github.com/eclipse/eclipse.jdt.ls/blob/d4101ae14289ff7db98763cf731a0c74138589b8/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BuildWorkspaceHandler.java#L100

I think we can use that directly for exporting jar task.

PR #1301, opened 2023-02-22
if current file is not a main class, 'Run Java' will run previously used app

### LINKED ISSUE #1036: Add a "launch previously launched application option if the current file isn't executable" option.
Add an option to launch the last application if the current one isn't executable instead of showing a dialog box. If there's a feature for this already, please do let me know.

testforstephen:
Fixes #1036

**Old**:
- If current file is a main class, run it directly.
- If current file is not a main class, it will always pop up a dialog to select the target main class to run.

**New**:
- If current file is a main class, "Run Java" will still run current file as past.
- If current file is not a main class, then run previously launched application automatically.

https://user-images.githubusercontent.com/14052197/220583881-62c6abb5-981b-4be6-ae0d-f53d98967181.mp4


src/extension.ts
```diff
@@ -398,3 +410,15 @@ async function runJavaProject(node: any, noDebug: boolean) {
         throw ex;
     }
 }
+
+function findLaunchConfiguration(mainClass: string, projectName: string | undefined, filePath?: string): vscode.DebugConfiguration | undefined {
+    const workspaceFolder: vscode.WorkspaceFolder | undefined =
+            filePath ? vscode.workspace.getWorkspaceFolder(vscode.Uri.file(filePath)) : undefined;
+    const launchConfigurations: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("launch", workspaceFolder);
+    const existingConfigs: vscode.DebugConfiguration[] = launchConfigurations.configurations;
+    const existConfig: vscode.DebugConfiguration | undefined = _.find(existingConfigs, (config) => {
```
jdneo, line 419: What if the user has two debug configurations, whose `mainClass` and `projectName` are the same but other options like args, vmArgs are different?

testforstephen: The first one will be used. This might be a rare case, since no complaint received yet.

jdneo (approved):

Issue #1342, opened 2023-05-11
No delegateCommandHandler for vscode.java.resolveMainMethod


SofijaErkin:
[VS Code has prompted "No delegateCommandHandler" while debugging the Java program]

##### Environment
- Operating System: `macOS Monterey Version 12.6.5`
- JDK version: `1.8.0_292 (x86_64) "AdoptOpenJDK"`
- Visual Studio Code version: `1.78.1`
- Java extension version: `(Language Support for Java(TM) by Red Hat) v0.64.1`
- Java Debugger extension version:` (Debugger for Java) v0.50.0`

##### Steps To Reproduce
1. Create a java file and the launch JSON file for VS Code, then compile and run.

Compile and run are successful, also, output the result of this java program. just liking:

<img width="867" alt="Screen Shot 2023-05-11 at 22 03 58" src="https://github.com/microsoft/vscode-java-debug/assets/28106861/25a6f77f-1d2e-4290-822d-57919481ee7f">

2. Run&Debug using "StartDebugger(F5)" Button

VS Code has prompted "`No delegateCommandHandler for vscode.java.resolveMainMethod`".

##### A Sample Project 

My workspace, just liking:

    testJavaDevelopDemo
    |____  .vscode
    |           |___ launch.json
    |____ testDemo.java
My "`testJavaDevelopDemo/.vscode/launch.json`" just liking:

    {
        // Use IntelliSense to learn about possible attributes.
        // Hover to view descriptions of existing attributes.
        // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
        "version": "0.2.0",
        "configurations": [
        
            {
                "type": "java",
                "name": "Debug (Launch)",
                "request": "launch",
                "mainClass": "${file}",
                "console": "integratedTerminal"
             }
       ]
    }

My "`testJavaDevelopDemo/testDemo.java`' just liking:

    public class testDemo {
        public static void main(String[] args) {
               int a=1;
               int b=2;
               int c = a + b;
               System.out.println(a+"+"+b+"="+c);
               System.out.println("hello world");
        }

    }

##### LOGS

Sorry, I had not known how to check out the logs of debugging.

##### Current Result

The `VSCode` has prompted "`No delegateCommandHandler for vscode.java.resolveMainMethod`"

while I click the "`Start Debugger(F5)`" button. Just like the below picture showing:

[...]

Issue #1308, opened 2023-03-09
No delegateCommandHandler for vscode.java.resolveMainMethod


tcranw:
When running any .java file from VS Code, it produces the error "No delegateCommandHandler for vscode.java.resolveMainMethod". I have tried restarting and reinstalling VS Code and there was no change.

- Operating System: Windows 11
- JDK version: 17.0.6
- Visual Studio Code version: 1.76.0
- Java extension version: Newest (reinstalled)
- Java Debugger extension version: Newest (reinstalled)

tcranw:
Downgrading versions of the Test Runner for Java extension fixed my issue

PR #1343, opened 2023-05-16
Add more solutions for No delegateCommandHandler via updating Troubleshooting.md

SofijaErkin:
I've just added two solutions about the Trouble, "`No delegateCommandHandler`", all the two solutions fixed the trouble. 
They are referring from the issues #1342 and #1308.

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -112,7 +112,11 @@ Cause of error is for now unknown, but something caused vscode java support and 
 
 ### Try:
 1. Restart VS Code and the issue should disappear 
-2. If it continues to error try restart again, and if still a problem open an issue at [vscode-java-debug](https://github.com/Microsoft/vscode-java-debug)
+2. If it continues to error try restart again
+3. If the version of Java Debugger extension and RedHat Java extension are not the most recent version, referring [#1342](https://github.com/microsoft/vscode-java-debug/issues/1342), please update both of them to the most recent version, and reboot VS Code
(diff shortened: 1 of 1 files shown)


Troubleshooting.md
```diff
@@ -112,7 +112,11 @@ Cause of error is for now unknown, but something caused vscode java support and
 
 ### Try:
 1. Restart VS Code and the issue should disappear 
-2. If it continues to error try restart again, and if still a problem open an issue at [vscode-java-debug](https://github.com/Microsoft/vscode-java-debug)
+2. If it continues to error try restart again
+3. If the version of Java Debugger extension and RedHat Java extension are not the most recent version, referring [#1342](https://github.com/microsoft/vscode-java-debug/issues/1342), please update both of them to the most recent version, and reboot VS Code
+4. If the version of Java Debugger extension and RedHat Java extension are the most recent version or even the newest, and if using the Test Runner for Java extension, referring [#1308](https://github.com/microsoft/vscode-java-debug/issues/1308), please downgrade the version of Test Runner for Java extension, then
```
jdneo, line 117: It doesn't make sense to me, personally.

If we can confirm that Test Runner may cause this issue, we should find the root cause and fix it. This statement does not provide enough information. Which version does user expect to downgrade?

testforstephen: +1, all extensions in vscode-java-pack should be compatible with their latest versions.

sanektmb (approved):

Vstuelke (approved):

Issue #1314, opened 2023-03-19
VSCode will not generate annotations when running tests


kjq:
VSCode does not generate annotation when "launch before build" or "building using the Java Projects" view.  

I have been having a problem with VSCode, in test cases, where my "annotations" are not getting generated _only_ when I run/launch before build a test case (or if i manually build the project using the Java Projects view).

**Scenarios**:
- Run a test case with "launch before build" - nothing is generated under generated-sources/annotations
- Run Clean Workspace from the Java Projects tree - nothing is generated under generated-sources/annotations
- Run Rebuild Project from the Java Projects - nothing is generated under generated-sources/annotations
- Run Rebuild All from the Java Projects - nothing is generated under generated-sources/annotations
- Run mvn from the command-line - everything works.

Even though VSCode will show 100's of errors the test still runs and can be debugged in the editor. Everything else also seems to work fine.

I've been trying to reproduce this but have not been able to with a small example yet.

Referencing this issue: `https://github.com/microsoft/vscode-java-debug/issues/1232`

testforstephen:
Since you're using mapstruct for annotation generation, could you take a look at the issue
https://github.com/redhat-developer/vscode-java/issues/3007?

kjq:
If I understand correctly, the solution above with MapStruct was to "move" the SPI piece into a separate Maven dependency.  We are not creating any special implementations around the SPI.  If the suggestion is to move the mappings themselves into another Maven dependency then that is not really possible since they sit, purposely, within the module using them.

Even though VSCode flag everything as errors I _can_ debug and run the code.

At the very least, it seems that I can no longer disable the `forceBuildBeforeLaunch` and I have made the change across _all_ of my settings as far as I know.  At least when that didnt build during testing I could build outside in the CLI and then debug without all the errors.

```
"java.autobuild.enabled": false,
"java.debug.settings.forceBuildBeforeLaunch": false,
```

Also, pretty sure, but it has been a while, that debugging "did work" with `foceBuildBeforeLaunch` as long as I compiled outside of VSCode _if_ the mappings were changed but everything else was fine.

The reason this is an issue is because we do not change the mappings that often but we do change other code and having to remember to build outside is sort of painful (but would be acceptable at this point).

testforstephen:
`forceBuildBeforeLaunch` only takes effect when the autobuild setting is enabled, and it is used to avoid duplicated building. But if autobuild is disabled, we still force a build before launching.

[...]

PR #1380, opened 2023-08-28
Respect forceBuildBeforeLaunch setting even if autobuild is disabled

testforstephen:
This allows user to use Maven CLI for build to workaround the issue https://github.com/microsoft/vscode-java-debug/issues/1314


jdneo (approved):
Assuming that user disables both `autobuild` and `forceBuildBeforeLaunch` is aware of that he should build on his own. I think it should be very small amout of people. This change looks good to me.

testforstephen:
> Assuming that user disables both `autobuild` and `forceBuildBeforeLaunch` is aware of that he should build on his own. I think it should be very small amout of people. This change looks good to me.

agree. The default value of both the settings are true. If users explicitly disable it, they should know what they do.

PR #1498, opened 2024-07-05
Add the doc for the launch config 'javaExec' in README

testforstephen:


jdneo (approved):
Consider to update https://code.visualstudio.com/docs/java/java-debugging#_launch as well

Issue #1505, opened 2024-08-22
Launcher to support multiple envFiles


hildo:
Currently, when launching a Java application, a single envFile is supported.  It would be nice of environment variables could be defined from multiple environment files.

This is supported in other vscode extensions (this is one for example, https://github.com/Microsoft/vscode-go/issues/1746)

##### Environment
- Operating System:
- JDK version:
- Visual Studio Code version:
- Java extension version:
- Java Debugger extension version:

##### Steps To Reproduce
1. Try to define an Array of envFile locations when launching a Java application

{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "java",
            "name": "May Application",
            "request": "launch",
            "cwd": "${workspaceFolder}",
            "mainClass": "org.csea.myapp.MyApplication",
            "projectName": "myapp",
            "envFile": [
            	"${workspaceFolder}/envs/.env1",
            	"${workspaceFolder}/envs/.env2"
    	    ]
        }

    ]
}

testforstephen:
this has been supported by https://github.com/microsoft/vscode-java-debug/pull/1506

hildo:
Thank you for that... what is the best way to get notified when a release of the plugin containing these changes will be available?

testforstephen:
The pre-release v0.58.2024090204 is just released. The stable release usually happens at the end of the month.

hildo:
Thank you sir!

PR #1506, opened 2024-08-22
Changes that will allow an array of envFile locations to be defined

hildo:
Changes for issue #1505

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Setup Node.js environment
       uses: actions/setup-node@v2
       with:
-        node-version: 16
+        node-version: 20
         
     - name: Install Node.js modules
       run: npm install
@@ -65,7 +65,7 @@ jobs:
     - name: Setup Node.js environment
       uses: actions/setup-node@v2
       with:
-        node-version: 16
+        node-version: 20
         
     - name: Install Node.js modules
       run: npm install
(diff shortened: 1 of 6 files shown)


hildo:
Is there something I can do to help here?  I don't understand why my changes would be causing these failures?  It looks like it is related to this step?
image

hildo:
It looks like the tool being used (vsce) needs NodeJS minimum version 20, and the build pipeline is using version 16?
image
Should Node be upgraded?

hildo:
@jdneo @testforstephen I have attempted to update the CI build to use NodeJS version 20, with the hope that the build issues will be addressed.  Could someone please approve the workflow so we can see if the changes were sufficient?  If I understand correctly, the workflow approval is not Pull Request approval, and just running the pipeline against this feature branch.

testforstephen:
please also update the envFile's type definition (should be `["array", "string"]`) in package.json and description in package.nls.json

https://github.com/microsoft/vscode-java-debug/blob/4a903c4c3c43e3072dc5381cd224170f37799631/package.json#L488-L489

And 
https://github.com/microsoft/vscode-java-debug/blob/4a903c4c3c43e3072dc5381cd224170f37799631/package.nls.json#L21

hildo:
Of course.  Thank you for the review.  Please let me know if I have done them correctly  (I am not 100% sure that my changes to package.json are correct).

testforstephen (approved):

hildo:
Thank you!

PR #1578, opened 2025-10-15
feat: no config debug

wenytang-ms:
Feature request https://github.com/microsoft/vscode-java-debug/issues/1530

This PR introduces a configuration-less debugging experience for Java applications, similar to the JavaScript Debug Terminal in VS Code. Users can now debug Java programs directly from the terminal without creating or maintaining launch.json files.

Usage Examples
Basic Usage
```
# Instead of: java -cp . com.example.Main
javadebug -cp . com.example.Main
```

With JAR Files
```
javadebug -jar target/myapp.jar
```

Spring Boot Applications
```
javadebug -jar myapp.jar --spring.profiles.active=dev
```

[Demo Record](https://microsoftapc-my.sharepoint.com/:v:/g/personal/wenyutang_microsoft_com/EZseyFht1h9BpBqY7HF3MmIBxKNGCkbYYDF6mBAAenBxAA?e=hpXJg0&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D)

PR #1648, opened 2026-06-02
fix: avoid spurious 'restart terminal' prompts on every window reload

### LINKED ISSUE #1647: Extension keeps requesting to restart my terminal
Every time I start or reload my window the extension wants to restart my terminal.
This is annoying.

- Version: 0.59.0

Steps to reproduce:

- Open Terminal in VSCode (Bash)
- `> Reload Window`
- Warning appears after a few seconds

Image

---

Another extension had the same issue and was able to resolve it (eventually):

- https://github.com/gitkraken/vscode-gitlens/issues/4977

wenytang-ms:
## Summary

Fixes #1647: the Debugger for Java extension prompts the user to restart their terminal every time the VS Code window is reloaded, even when nothing about the contributed environment has actually changed.

## Root cause

 + "src/noConfigDebugInit.ts" +  runs on every activation and unconditionally:

1. Calls  + "collection.clear()" +  on  + "context.environmentVariableCollection" + .
2. Re-adds  + "VSCODE_JDWP_ADAPTER_ENDPOINTS" + ,  + "VSCODE_JAVA_EXEC" + , and a  + "PATH" +  append.

 + "EnvironmentVariableCollection" +  is persistent ( + "persistent = true" + ) and VS Code triggers the `restart terminal` prompt whenever the collection differs from what was last applied to running terminals. `clear()` followed by a re-add is always interpreted as a change, even when the resulting set of variables is byte-identical, so the prompt fires on every reload (see microsoft/vscode#95919, microsoft/vscode#234384).

## Fix

Introduce diff-aware helpers in `src/envVarSync.ts`:

- `applyReplaceIfChanged` / `applyAppendIfChanged` — only call `replace` / `append` when the existing mutator's type, value, **and normalized options** differ from the desired ones.
- `deleteIfPresent` — only call `delete` when a mutator exists.

Rewrite the activation flow in `src/noConfigDebugInit.ts` to use those helpers and drop the unconditional `clear()`. On a typical reload all comparisons are no-ops, so VS Code never observes a change and never prompts.

### Behavioral notes

- **Java executable detection failures no longer wipe `VSCODE_JAVA_EXEC`.** If `getJavaHome()` returns nothing (e.g., JDT LS hasn't finished initializing), we keep the previously stored value instead of deleting it, avoiding churn from transient startup hiccups.
- **One-time legacy cleanup**: explicitly `delete` keys older versions used to set but no longer manage (currently just `JAVA_TOOL_OPTIONS`, per the comment that was already in this file). After the first reload this is also a no-op.
- **Description**: set `collection.description = `Java No-Config Debug`` so the source is visible in VS Code's environment-variable UI.

## Tests

Added `test/envVarSync.test.ts` with 13 cases covering:

- write-on-missing, no-op-on-match, write-on-value-diff, write-on-options-diff
- override across mutator types (Replace ↔ Append)
- `deleteIfPresent` present/absent
- a regression test simulating two back-to-back activations and asserting the second one performs **zero** writes — this is the exact scenario from #1647.

Full `npm test` output: **24 passing, 0 failing**.

## Risk

- Scope is limited to the No-Config Debug activation path.
- No public API surface changes; `registerNoConfigDebug` signature unchanged.
- One-time prompt may still appear after upgrading to a version whose `extPath` differs (the `PATH` append value embeds the extension version directory) — this is unavoidable and only happens once per upgrade.

### DIFF
--- a/src/noConfigDebugInit.ts
+++ b/src/noConfigDebugInit.ts
@@ -9,12 +9,7 @@ import * as vscode from 'vscode';
 import { sendInfo, sendError } from "vscode-extension-telemetry-wrapper";
 import { getJavaHome } from "./utility";
 import { buildNoConfigPathAppendValue } from "./pathUtil";
-import { applyAppendIfChanged, applyReplaceIfChanged, deleteIfPresent } from "./envVarSync";
-
-// Environment variables that older versions of this extension contributed but
-// no longer manage. They are cleaned up explicitly on activation so they do
-// not linger in the persistent EnvironmentVariableCollection. See issue #1647.
-const LEGACY_ENV_VARS = ["JAVA_TOOL_OPTIONS"];
+import { applyAppendIfChanged, applyReplaceIfChanged } from "./envVarSync";
 
 const ENV_VAR_COLLECTION_DESCRIPTION = "Java No-Config Debug";
 
@@ -81,13 +76,6 @@ export async function registerNoConfigDebug(
     // see which extension is contributing these variables.
     if (collection.description !== ENV_VAR_COLLECTION_DESCRIPTION) {
         collection.description = ENV_VAR_COLLECTION_DESCRIPTION;
-    }
-
-    // Remove any variables that older versions of this extension used to set
-    // but no longer manage. Doing this idempotently (only when present) keeps
-    // subsequent reloads from triggering the "restart terminal" prompt.
-    for (const legacy of LEGACY_ENV_VARS) {
-        deleteIfPresent(collection, legacy);
     }
 
     // Apply our managed variables using diff-aware helpers. On a typical
--- a/src/noConfigDebugInit.ts
+++ b/src/noConfigDebugInit.ts
@@ -9,6 +9,14 @@ import * as vscode from 'vscode';
 import { sendInfo, sendError } from "vscode-extension-telemetry-wrapper";
 import { getJavaHome } from "./utility";
 import { buildNoConfigPathAppendValue } from "./pathUtil";
+import { applyAppendIfChanged, applyReplaceIfChanged, deleteIfPresent } from "./envVarSync";
+
+// Environment variables that older versions of this extension contributed but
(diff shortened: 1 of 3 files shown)


src/noConfigDebugInit.ts
```diff
@@ -69,30 +77,46 @@ export async function registerNoConfigDebug(
...
+    // Surface a description in VS Code's environment variable UI so users can
+    // see which extension is contributing these variables.
+    if (collection.description !== ENV_VAR_COLLECTION_DESCRIPTION) {
+        collection.description = ENV_VAR_COLLECTION_DESCRIPTION;
+    }
+
+    // Remove any variables that older versions of this extension used to set
+    // but no longer manage. Doing this idempotently (only when present) keeps
+    // subsequent reloads from triggering the "restart terminal" prompt.
+    for (const legacy of LEGACY_ENV_VARS) {
```
chagong: Is this kind of dangerous? if it's legacy var, we can just leave it. What if users add this env var manually?

wenytang-ms:
Good catch — agreed, removed in bf7fff1.

To clarify why it was safe but still worth removing:

- `context.environmentVariableCollection` is **per-extension**. `collection.delete()` only removes a mutator that *this extension* previously wrote; it can never touch a `JAVA_TOOL_OPTIONS` the user set in their shell profile, `settings.json terminal.integrated.env.*`, or another extension's collection.
- I checked the history with `git log -S JAVA_TOOL_OPTIONS -- src/noConfigDebugInit.ts` — this extension has **never** called `collection.replace('JAVA_TOOL_OPTIONS', ...)`. The only reference is the comment from #1578 explaining why we deliberately don't set it. So the cleanup loop was dead code.

Since the value was effectively zero and the code reads as if it might be doing something risky, dropping it is the cleaner call. PR updated.

src/noConfigDebugInit.ts
wenytang-ms: deleted

chagong (approved):

PR #1649, opened 2026-06-03
Allow Exact Match For Filters

tthornton3-chwy:
I was having cases where filters were matching inappropriately, for example with composite builds. This lets you have an exact match syntax. I know it's a bit niche, but I do think it's useful/I personally use it!


tthornton3-chwy:
@microsoft-github-policy-service agree

chagong:
Thanks, this is a reasonable addition and the trailing-separator approach is backward compatible. One blocker and two smaller asks before merge:

**Windows separator gap.** The typical config form is `!${workspaceFolder}/build/classes/` — after resolution that's an absolute path with forward slashes and a trailing `/`. Since it's absolute, `path.join` is skipped, so on Windows `exclude.endsWith('\\')` is `false` and exact-match silently never triggers. A Windows user would have to hand-write backslashes. Suggest making the check separator-agnostic and running it before normalization:

```ts
if (/[\\/]$/.test(exclude)) {
    exclude = exclude.slice(0, -1);
    isDirect = true;
} else {
    isDirect = this.isFilePath(exclude);
}
```

**Docs.** The trailing-slash convention isn't discoverable — can you update `java.debugger.launch.classPaths.exclude` and `modulePaths.exclude` in `package.nls.json` to mention it? It's also slightly counter-intuitive (a trailing slash usually implies a directory/prefix, here it means exact match), so a clear note helps.

**Test.** A small unit test covering the new exact-match path in `filterExcluded` would be good given this logic is filesystem-sensitive and currently untested.

tthornton3-chwy:
@chagong I finally got around to addressing the (very good!) feedback :) Can you re-take a peek?

src/configurationProvider.ts
```diff
@@ -490,25 +490,35 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
         const excludes: Map<string, boolean> = new Map<string, boolean>();
         for (const p of paths) {
             if (p.startsWith("!")) {
-                let exclude = p.substr(1);
+                let exclude = p.slice(1);
+                let isDirect: boolean;
+
+                if (/[\\/]$/.test(exclude)) {
```
tthornton3-chwy, line 496: before normalization, and using the exact code-block suggested, per feedback!

src/configurationProvider.ts
```diff
@@ -490,25 +490,35 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
             }
 
             result.push(vscode.Uri.file(p).fsPath);
         }
 
         return result.filter((r) => {
-            for (const [excludedPath, isFile] of excludes.entries()) {
-                if (isFile && r === excludedPath) {
+            for (const [excludedPath, isDirect] of excludes.entries()) {
+                if (isDirect && r === excludedPath) {
```
chagong: Could you normalize trailing separators consistently on both sides before this exact comparison? Currently, `["C:\\workspace\\lib\\", "!C:\\workspace\\lib\\"]` leaves the separator on `r` but removes it from `excludedPath`, so the directory is not excluded. Please also add forward-slash and backslash tests where the included entry ends in a separator, taking care not to alter filesystem roots.

package.nls.json
```diff
@@ -8,12 +8,12 @@
     "java.debugger.launch.modulePaths.auto": "Automatically resolve the module paths of current project.",
     "java.debugger.launch.modulePaths.runtime": "The module paths within 'runtime' scope of current project.",
     "java.debugger.launch.modulePaths.test": "The module paths within 'test' scope of current project.",
-    "java.debugger.launch.modulePaths.exclude": "The path after '!' will be excluded from the modulePaths.",
+    "java.debugger.launch.modulePaths.exclude": "The path after '!' will be excluded from the modulePaths. A slash (forwards or backwards) will treat the path as an exact match.",
```
chagong: This should say “A trailing slash or backslash will treat the path as an exact match.” The current wording suggests that any slash triggers exact matching, while the implementation only checks the final character. Please make the same change to the `classPaths.exclude` description.

chagong (changes requested):
The revised Windows separator handling and tests address the earlier feedback, but the trailing-separator comparison and documentation wording still need adjustment before merge.

tthornton3-chwy:
@chagong I believe i have addressed round 2 feedback! Thanks again!

src/configurationProvider.ts
```diff
@@ -490,25 +490,35 @@ export class JavaDebugConfigurationProvider implements vscode.DebugConfiguration
...
+                    isDirect = this.isFilePath(exclude);
+                }
+
                 if (!path.isAbsolute(exclude)) {
                     exclude = path.join(folder?.uri.fsPath || "", exclude);
                 }
+
                 // use Uri to normalize the fs path
-                excludes.set(vscode.Uri.file(exclude).fsPath, this.isFilePath(exclude));
+                excludes.set(vscode.Uri.file(exclude).fsPath, isDirect);
```
chagong, line 508: Stripping and classifying `exclude` before workspace resolution causes two regressions: filesystem roots such as `C:\` lose their root separator, and relative files such as `!lib/foo.jar` are checked against the extension host's working directory and become prefix exclusions. We can keep the separator, record whether it selected exact matching, and classify files only after resolving and normalizing the path:

```suggestion
                let exclude = p.slice(1);
                const isExactMatch = /[\\/]$/.test(exclude);

                if (!path.isAbsolute(exclude)) {
                    exclude = path.join(folder?.uri.fsPath || "", exclude);
                }

                // use Uri to normalize the fs path
                exclude = vscode.Uri.file(exclude).fsPath;
                excludes.set(exclude, isExactMatch || this.isFilePath(exclude));
```

This preserves roots while the existing comparison still strips non-root trailing separators from both operands. Could you also add regression tests for `[root, `!${root}`]` and for a relative file exclusion retaining a sibling such as `foo.jar.backup`?

